PR #1, opened 2019-08-26
Adding boot order manipulation and few other changes

smalleni:


tasks/setup_infrared.yml
```diff
@@ -3,6 +3,8 @@
   git:
       repo: "https://github.com/redhat-openstack/infrared"
       dest: "{{ infrared_dir }}"
+      force: yes
```
venkataanil, line 6: We don't want to forcefully clone if it exists.

group_vars/all.yml
```diff
@@ -0,0 +1,35 @@
...
+controller_count: 3
+# No need to set compute_count. This will be set to all remaining nodes, which is calclulated in overcloud.yml
+#compute_count: 1
+alias:
+#lab specific vars
+  lab_url: "http://quads.alias.bos.scalelab.redhat.com"
+scale:
+# lab specific vars
+  lab_url: "http://quads.rdu2.scalelab.redhat.com"
+machine_types:
```
venkataanil, line 22: We can get vendor name by reading sys entries
[root@f16-h10-000-1029p ~]# cat /sys/class/dmi/id/board_vendor
Supermicro
[root@e23-h33-740xd ~]# cat /sys/class/dmi/id/board_vendor
Dell Inc.
[vkommadi@localhost ~]$ cat /sys/class/dmi/id/board_vendor
LENOVO

But this change is more efficient which allows to find the vendor without login into the machine.

venkataanil (commented):
Thanks for the patch. Want to merge if you remove the "force" option for cloning infrared.

group_vars/all.yml
venkataanil: I mean your change (i.e getting vendor by parsing url) is more efficient :)

smalleni:
Closing as no longer relevant. Future commits will address this.

PR #20, opened 2019-09-19
Add sshpass

zulcss:
sshpass is required on the localhost, install it if its not installed
already.


smalleni:
Good for now! In the future we might want to combine all the install tasks and loop over them.

smalleni (approved):

PR #25, opened 2019-09-23
Dynamic provision of OS based on OSP release

smalleni:
This commit adds support to dynamically provision undercloud
OS depending on the version of OSP required.  Earlier only RHEL 8 install
was supported. Also, now the provisioning only takes place if there is a
mismatch between OS version installed and that required by OSP.

### DIFF
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -59,63 +59,102 @@
       set_fact:
         chassis_password:  "{{ instackenv_content.nodes[0].pm_password }}"
 
-    - name: RHEL 8 install for OSP 15 and above
+    - name: Check if /etc/redhat-release exists
+      stat:
+       path: /etc/redhat-release
+      register: rhel_release_stat
+      delegate_to: "{{ undercloud_hostname }}"
+
+    - name: Get RHEL version
+      become: true
+      slurp:
+        src: /etc/redhat-release
+      register: file_content
+      when: rhel_release_stat.stat.exists
+      delegate_to: "{{ undercloud_hostname }}"
+
+    - name: set version fact
+      set_fact:
+        version: "{{ file_content.content | b64decode }}"
+
+    - name: set rhel version
+      set_fact:
+        rhel_version: "{{ version.split()[-2] }}"
+
+    - name: RHEL 8 install if needed
+      shell: hammer -u  {{ cloud_name }} -p {{ chassis_password }}  host update --name {{ undercloud_hostname }} --build 1 --operatingsystem 'RHEL 8' --overwrite 1
+      delegate_to: "{{ hammer_host }}"
+      register: rhel8_install
+      when: osp_release | int > 14 and  rhel_version is version('8.0', '<')
+
+    - name: RHEL 7 install if needed
+      shell: hammer -u {{ cloud_name }} -p {{ chassis_password }}  host update --name {{ undercloud_hostname }} --build 1 --operatingsystem 'RHEL 7' --overwrite 1
+      delegate_to: "{{ hammer_host }}"
+      register: rhel7_install
+      when: osp_release | int < 15 and  rhel_version is  version ('8.0', '>=')
+
+    - name: Reboot if OS install needed
       block:
-          - name: Set foreman RHEL 8 build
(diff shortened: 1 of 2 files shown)


smalleni:
@venkataanil PTAL

setup_undercloud.yml
```diff
@@ -60,63 +60,93 @@
...
+      delegate_to: "{{ undercloud_hostname }}"
+
+    - name: set version fact
+      set_fact:
+        version: "{{ file_content.content | b64decode }}"
+
+    - name: set rhel version
+      set_fact:
+        rhel_version: "{{ version.split()[-2] }}"
+
```
zulcss, line 83: We can get this from infrared itself cant we?

smalleni: This is checking for the already installed RHEL version on the host we want to install Undercloud on. I don’t see how we can get it from infrared. Thoughts ?

setup_undercloud.yml
```diff
@@ -60,63 +60,93 @@
...
+          retries: 100
+          until: uc_reachable|succeeded
+
+        - name: add keys to new undercloud
+          include_tasks: tasks/copykeys.yml
+          vars:
+            hostname: "{{ undercloud_hostname }}"
+      when: rhel7_install is not skipped or rhel8_install is not skipped
+
+    - name: add undercloud_host to inventory file
```
venkataanil: This task has to be run with "when: rhel7_install is not skipped or rhel8_install is not skipped"
As you know, you have to change ansible_python_interpreter depending on rhel7 or 8.

setup_undercloud.yml
```diff
@@ -60,63 +60,93 @@
...
+        rhel_version: "{{ version.split()[-2] }}"
+
+    - name: RHEL 8 install if needed
+      shell: hammer -u  {{ cloud_name }} -p {{ chassis_password }}  host update --name {{ undercloud_hostname }} --build 1 --operatingsystem 'RHEL 8' --overwrite 1
+      delegate_to: "{{ hammer_host }}"
+      register: rhel8_install
+      when: osp_release | int > 14 and  rhel_version is version('8.0', '<')
+
+    - name: RHEL 7 install if needed
+      shell: hammer -u {{ cloud_name }} -p {{ chassis_password }}  host update --name {{ undercloud_hostname }} --build 1 --operatingsystem 'RHEL 8' --overwrite 1
```
venkataanil: typo, mistakenly added 'RHEL 8' instead of 'RHEL 7' :)

venkataanil (commented):
need to test :)

setup_undercloud.yml
smalleni: @venkataanil So, I add the host to inventory only if RHEL 8 was installed and you can see in line 148 I set the python interpreter. Even in prepare_inventory.yml I add the python interprete if host is RHEL 8.

smalleni:
@venkataanil PTAL

tasks/prepare_inventory.yml
```diff
@@ -1,9 +1,25 @@
...
 - name: add localhost to inventory file
   add_host:
-      name: "localhost"
-      groups: "local"
-      ansible_connection: "local"
+    name: "localhost"
+    groups: "local"
+    ansible_connection: "local"
+
+- name: Ping undercloud and set python version for RHEL 8
```
venkataanil, line 8: nice idea for finding OS version without logging into the machine.

setup_undercloud.yml
```diff
@@ -59,63 +59,102 @@
...
+      when: rhel8_install is not skipped
+
+    - name: add undercloud_host to inventory file (RHEL 7)
+      add_host:
+        name: "{{ undercloud_hostname }}"
+        groups: "baremetal,undercloud,tester"
+        ansible_host: "{{ undercloud_hostname }}"
+        ansible_user: "{{ ansible_ssh_user }}"
+        ansible_ssh_private_key_file: "{{ ansible_ssh_key }}"
+      when: rhel8_install is not skipped
```
venkataanil: typo, patch is good to go after fixing this typo.

venkataanil (commented):
Please fix the typo. NIce approach to fix OS versions.
We can merge after fixing the typo.

setup_undercloud.yml
smalleni: Good catch.

smalleni: @venkataanil fixed

PR #40, opened 2019-10-01
Let infrared manage yum repos

zulcss:
Before we were setting the osp repos manually, but let infrared do it
for us.


smalleni:
@zulcss merge conflicts?

PR #45, opened 2019-10-02
Fix package install on fedora30

zulcss:
python-virtualenv and libselinux-python doesnt exist on f30,
Add python3-virtualenv and python3-libselinux which do.

### DIFF
--- a/tasks/setup_infrared.yml
+++ b/tasks/setup_infrared.yml
@@ -9,7 +9,7 @@
       - libffi-devel
       - python-virtualenv
       - libselinux-python
-      - python30-virtualenv
+      - python3-virtualenv
       - python3-libselinux
   become: true
   failed_when: false
--- a/tasks/setup_infrared.yml
+++ b/tasks/setup_infrared.yml
@@ -9,7 +9,10 @@
       - libffi-devel
       - python-virtualenv
       - libselinux-python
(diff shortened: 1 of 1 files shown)


tasks/setup_infrared.yml
```diff
@@ -9,7 +9,10 @@
       - libffi-devel
       - python-virtualenv
       - libselinux-python
+      - python30-virtualenv
+      - python3-libselinux
   become: true
+  failed_when: false
```
smalleni, line 15: Why is it OK to skip package installs?

tasks/setup_infrared.yml
```diff
@@ -9,7 +9,10 @@
       - libffi-devel
       - python-virtualenv
       - libselinux-python
+      - python30-virtualenv
```
smalleni: python30?

smalleni (changes requested):

tasks/setup_infrared.yml
venkataanil: For now this will be fine. We will enhance it later

tasks/setup_infrared.yml
venkataanil: I will correct it.

venkataanil (approved):

PR #47, opened 2019-10-03
Retry pxe boot

venkataanil:
1) create badfish virtual environment and run badfish
from there
2) retry setting boot mode to pxe. Sometimes this will
fail, so rety can help

### DIFF
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -1,4 +1,6 @@
 - hosts: localhost
+  vars:
+      badfish_venv: "{{ ansible_user_dir }}/badfish/.venv"
   tasks:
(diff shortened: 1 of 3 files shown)


setup_undercloud.yml
```diff
@@ -9,17 +11,13 @@
         repo: "https://github.com/redhat-performance/badfish.git"
         dest: "{{ ansible_user_dir }}/badfish"
 
-    - name: Install python3 pyyaml
-      pip:
-        name: pyyaml
```
venkataanil, line 14: pyyaml is inside badfish/requirements.txt. So no need to install here

PR #49, opened 2019-10-04
external network access from undercloud

venkataanil:
external network access from undrecloud is needed to access
overcloud resources. We need to identify physical interface
which we added to br-ex in controller nodes.

We follow this strategy
1) When controller has seperate nic for external network,
we use the last available interface for external network.
four-nics-vlans config files use this approach.
2) When external network and control plane network share
the same physical nic, we use control plane interface as
external network interface.
three-nics-vlans use this approach.
After deploying overcloud we create a vlan interface on
this external interface and assign ip address to it.

### DIFF
--- a/files/virt/network/three-nics-vlans/compute.yaml.j2
+++ b/files/virt/network/three-nics-vlans/compute.yaml.j2
@@ -180,7 +180,7 @@ resources:
                           members:
                             -
                               type: interface
-                              name: nic1
+                              name: ctlplane_interface
                               primary: true
                           use_dhcp: false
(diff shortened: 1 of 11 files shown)


smalleni:
CI is failing..

zulcss (commented):
I would like to see the vlan setup also done in the /etc/sysconfig/network-scripts since if the user reboots the undercloud then that interface is gone.

venkataanil:
I will propose a separate PR for adding interface details in /etc/sysconfig/network-scripts

PR #53, opened 2019-10-07
Use template to configure undercloud.conf

zulcss:
Rather than using the CLI to configure the undercloud.conf
use a template.

### DIFF
--- a/templates/undercloud.conf.j2
+++ b/templates/undercloud.conf.j2
@@ -1,0 +1,19 @@
+[DEFAULT]
+local_interface = {{ ctlplane_interface }}
+local_ip = {{ gateway }}/{{ subnet_mask }}
+undercloud_ntp_servers = clock.redhat.com,clock2.redhat.com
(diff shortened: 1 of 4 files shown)


templates/undercloud.conf.j2
```diff
@@ -0,0 +1,21 @@
+[DEFAULT]
+
+local_interface = {{ ctlplane_interface }}
+local_ip = 192.168.24.1/24
+{% if osp_release <= 12 %}
+network_gateway = 192.168.24.1
```
venkataanil: Take the variables from group_vars as the same will be used in nic-configs. So IP addresses used in nic-configs and undercloud conf should match.

venkataanil (commented):
Thanks for the patch. Why can't we have 1092.168.24.0/24 network for OSP10?

venkataanil (approved):
Tested on OSP10 and OSP13 and able to deploy undercloud.

PR #57, opened 2019-10-09
change undercloud_instackenv to overcloud_instackenv

asyedham:
This naming convention will help in identifying that the undercloud node is excluded from the instackenv.json file and new file  overcloud_instackenv.json is created.


venkataanil (approved):
Thanks

PR #63, opened 2019-10-09
Configure overcloud when the deployment finishes

zulcss:
- Add flavors
- Add images
- Configure quotas
- Add keypair
- Configure security groups


venkataanil (commented):
https://github.com/redhat-performance/jetpack/blob/master/browbeat.yml#L35
will create flavours, images.

zulcss:
I know, but I dont necessarily want to install browbeat.

masco:

PR #68, opened 2019-10-10
Add missing subnet mask

zulcss:
Add missing subnet_mask in group_vars/all.yml


zulcss:
This is fixed a different way

PR #71, opened 2019-10-10
Add a log directory

asyedham:


smalleni:
Can you squash these commits into one please

smalleni (commented):
Is there any reason to use &> {{ log_directory }}/overcloud_deploy.log

Maybe better to use > {{ log_directory }}/overcloud_deploy.log 2&>1

so that stderr is also redirected. 

Same for other commands....

PR #74, opened 2019-10-14
Fix up templates for OSP10

zulcss:
- Fix bad formatting in controller.yaml.j2
- Remove nic4 entry since its not being used.


venkataanil (approved):
Thanks for the patch

PR #84, opened 2019-10-24
[WIP]: composable roles

venkataanil:
Do not merge


venkataanil (commented):
No harm in merging changes. It can help if someone wants to work on top of it.
Only pending stuff is generating nic-configs for the composable roles.
As we are not calling this playbook, it won't break existing jetpack.

Issue #90, opened 2019-10-30
Installation fails has ssh keys are not generated


asyedham:
root@f16-h10-000-1029p jetpack]# ansible-playbook main.yml 
TASK [copy jump host key] *********************************************************************************************
fatal: [localhost]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/libexec/platform-python"}, "changed": false, "cmd": "echo '100yard-' | sshpass ssh-copy-id  -o 'StrictHostKeyChecking no' -f root@f16-h11-000-1029p.rdu2.scalelab.redhat.com\n", "delta": "0:00:00.016193", "end": "2019-10-30 12:31:01.146791", "msg": "non-zero return code", "rc": 1, "start": "2019-10-30 12:31:01.130598", "stderr": "/usr/bin/ssh-copy-id: ERROR: No identities found", "stderr_lines": ["/usr/bin/ssh-copy-id: ERROR: No identities found"], "stdout": "", "stdout_lines": []}
...ignoring

TASK [remove key and add again] ***************************************************************************************
fatal: [localhost]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/libexec/platform-python"}, "changed": false, "cmd": "ssh-keygen \necho '100yard-' | sshpass ssh-copy-id  -o 'StrictHostKeyChecking no' -f root@f16-h11-000-1029p.rdu2.scalelab.redhat.com\n", "delta": "0:00:00.057128", "end": "2019-10-30 12:31:01.430987", "msg": "non-zero return code", "rc": 1, "start": "2019-10-30 12:31:01.373859", "stderr": "/usr/bin/ssh-copy-id: ERROR: No identities found", "stderr_lines": ["/usr/bin/ssh-copy-id: ERROR: No identities found"], "stdout": "Generating public/private rsa key pair.\nEnter file in which to save the key (/root/.ssh/id_rsa): ", "stdout_lines": ["Generating public/private rsa key pair.", "Enter file in which to save the key (/root/.ssh/id_rsa): "]}

PLAY RECAP ************************************************************************************************************
localhost                  : ok=16   changed=2    unreachable=0    failed=1    skipped=0    rescued=0    ignored=1   

On a fresh RHEL8 setup, the installation fails if the ssh key is not generated on the ansible control node.

PR #91, opened 2019-10-30
 generates the ssh-key

asyedham:
Fixes the issue #90

### DIFF
--- a/tasks/copykeys.yml
+++ b/tasks/copykeys.yml
@@ -8,8 +8,8 @@
 
 - name: remove key and add again
   shell: |
-    ssh-keygen -R {{ undercloud_hostname }}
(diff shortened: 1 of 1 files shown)


asyedham:
@sai could you please have look

tasks/copykeys.yml
```diff
@@ -8,8 +8,8 @@
 
 - name: remove key and add again
   shell: |
-    ssh-keygen -R {{ undercloud_hostname }}
-    echo '{{ ansible_ssh_pass }}' | sshpass ssh-copy-id  -o 'StrictHostKeyChecking no' -f {{ ssh_user }}@{{ hostname }}
+    ssh-keygen -q -b 2048 -t rsa -N "" -C "creating SSH" -f {{ ansible_ssh_key }}
```
smalleni, line 11: Check if {{ ansible_ssh_key }} exists and only generate if needed.

smalleni (changes requested):

tasks/copykeys.yml
venkataanil: Agree

PR #94, opened 2019-11-04
Fix warnings

asyedham:
- Fix [WARNING]: Consider using 'become', 'become_method', and 'become_user' rather than running sudo
- fix lint


asyedham:
@venkataanil could you please have look

venkataanil (approved):

PR #96, opened 2019-11-05
Fix infrared plugin add all

### LINKED ISSUE #95: Installation fails due to missing plugin tripleo-undercloud
When the undercloud installation is attempted, we see an error due to missing tripleo-undercloud plugin. Looks like infrared plugin add all is failing at https://github.com/redhat-performance/jetpack/blob/master/tasks/setup_infrared.yml#L16. It might help to pip3 install virtualenv and setup the venv as python3 explicitly

smalleni:
Closes #95

### DIFF
--- a/tasks/setup_infrared.yml
+++ b/tasks/setup_infrared.yml
@@ -8,7 +8,8 @@
 - name: Add all plugins
   shell: |
       git fetch "https://review.gerrithub.io/redhat-openstack/infrared" refs/changes/76/463576/2 && git checkout FETCH_HEAD
-      virtualenv .venv && source .venv/bin/activate
+      pip3 install virtualenv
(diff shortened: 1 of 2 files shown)


tasks/setup_infrared.yml
```diff
@@ -8,7 +8,8 @@
 - name: Add all plugins
   shell: |
       git fetch "https://review.gerrithub.io/redhat-openstack/infrared" refs/changes/76/463576/2 && git checkout FETCH_HEAD
-      virtualenv .venv && source .venv/bin/activate
+      pip3 install virtualenv
```
venkataanil, line 11: What if ansible host is python 2.7?

smalleni: We have stated in our minimum requirements that we need python 3 right?

PR #97, opened 2019-11-05
OSP15 changes

venkataanil:


zulcss (commented):
This needs to be rediffed because of conflicts

PR #117, opened 2019-11-15
Use hammer host for all provision tasks

venkataanil:
We were setting up badfish in anisble jump host
to issue provision commands for rhel machines.
Instead we use hammer host for the same tasks.
This avoids installing badfish in ansible jump host.
note: badfish is already installed in hammer host
at /root/jetpack/abdfish.

### DIFF
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -1,7 +1,6 @@
 - hosts: localhost
   gather_facts: yes
   vars:
-      badfish_venv: "{{ ansible_user_dir }}/badfish/.venv"
       chassis_password:  "{{ instackenv_content.nodes[0].pm_password }}"
   tasks:
     - name: check if os installed as root user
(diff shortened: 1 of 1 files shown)


setup_undercloud.yml
```diff
@@ -39,24 +38,6 @@
...
-
-        - name: Create badfish virtualenv
-          command: virtualenv {{ badfish_venv }} creates={{ badfish_venv }}
-
-        - name: Install badfish requirements
-          pip:
-            requirements: "{{ ansible_user_dir }}/badfish/requirements.txt"
-            virtualenv: "{{ badfish_venv }}"
-
-        - include_tasks: tasks/copykeys.yml
```
venkataanil: Note: We are already adding keys at
https://github.com/redhat-performance/jetpack/blob/master/setup_undercloud.yml#L37
i.e this task is duplicated at L37 and L55.
So removing here i.e
https://github.com/redhat-performance/jetpack/blob/master/setup_undercloud.yml#L55

setup_undercloud.yml
```diff
@@ -67,30 +48,46 @@
         - name: set undercloud to PXE boot (Supermicro)
           shell: ipmitool -I lanplus -H mgmt-{{ undercloud_hostname }} -U quads -P {{ chassis_password }} chassis bootdev pxe options=persistent
           when: lab_name == "scale" and undercloud_hostname.split('.')[0].split('-')[3] in scale.machine_types.supermicro
+          delegate_to: "{{ hammer_host }}"
+          vars:
+            ansible_user: root
+            ansible_python_interpreter: "{{ hammer_python_interpreter }}"
```
asyedham: As the task to get the python interpreter is removed in the latest changes. hammer_python_interpreter var is undefined

asyedham: As the task to get the python interpreter is removed in the latest changes https://github.com/redhat-performance/jetpack/commit/573f116bb4f5e3e075cc198900c03751d1ac596e#diff-42149ab8289698b9e7c0c8490f19e6e1L63.hammer_python_interpreter var is undefined

venkataanil: Thanks Asma, fixed it now.

PR #121, opened 2019-11-16
Set timeout for introspection

smalleni:
Without this if a node does not introspect, jetpack is stuck at overcloud node instrospection until SSH timeout.


venkataanil:
Can we pass timeout option to infrared?

Issue #93, opened 2019-11-04
Need to remove epel releases in undercloud


venkataanil:
" yum install -y python-tripleoclient" is failing with below errors.

Total size: 110 M
Installed size: 445 M
Downloading packages:
warning: /var/cache/yum/x86_64/7Server/epel/packages/python2-iso8601-0.1.11-1.el7.noarch.rpm: Header V3 RSA/SHA256 Signature, key ID 352c64e5: NOKEY
Retrieving key from file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7

GPG key retrieval failed: [Errno 14] curl#37 - "Couldn't open file /etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7"

As we have epel releases this is happening. We need to remove epel before installing osp repos. i.e
rm -f /etc/yum.repos.d/epel*
rm -fr /var/cache/yum/*
yum clean all

PR #132, opened 2019-11-20
Remove default epel install

smalleni:
Cloeses #93


masco:

PR #135, opened 2019-11-27
allow user to force re provision the undercloud

masco:


setup_undercloud.yml
```diff
@@ -31,9 +31,15 @@
     - name: set os_install
       set_fact:
         os_install: "{{ (((osp_release|int > 14) and (rhel_version is version('8.0', '<')))|ternary('RHEL 8', ((osp_release|int < 15) and (rhel_version is version ('8.0', '>=')))|ternary('RHEL 7', false))) }}"
+      when: os_install == false
 
     - name: Reboot if OS install needed
       block:
+        - name: delete overcloud before re install undercloud
+          shell: |
+            ssh -o 'PreferredAuthentications=publickey' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -i ~/.ssh/id_rsa stack@{{ undercloud_hostname }} 'source ~/stackrc && openstack overcloud delete overcloud -y'
```
smalleni, line 45: You might want to add retries/delay and also add --wait option if you really want to delete overcloud. However it's not really a necessity, we can just reimage undercloud and deploy as the overcloud nodes are cleaned during enrollment and the OSes are wiped

setup_undercloud.yml
```diff
@@ -31,9 +31,15 @@
     - name: set os_install
       set_fact:
         os_install: "{{ (((osp_release|int > 14) and (rhel_version is version('8.0', '<')))|ternary('RHEL 8', ((osp_release|int < 15) and (rhel_version is version ('8.0', '>=')))|ternary('RHEL 7', false))) }}"
+      when: os_install == false
```
smalleni: So if user sets os_install to true because he wants to reimage undercloud, what version of OS is used to reimage (as it depends on os_install variable in https://github.com/redhat-performance/jetpack/blob/master/setup_undercloud.yml#L61). Maybe, best to use another variable like force_reprovision

smalleni (changes requested):

setup_undercloud.yml
masco: I will change this as you suggested for better clarity to the users.

setup_undercloud.yml
masco: I don't think we need to add any wait here, just checked this, the next task is executing once the command is completed.
regarding the retry, actually the provision is working without this task so I feel just give one try irrespective of the result we will continue with the provision. what do you think? I am open to change also :-)

venkataanil: May be osp10 is not having "openstack overcloud delete" option?

masco: @smalleni @venkataanil How about removing this task?

venkataanil (approved):

venkataanil (approved):

PR #139, opened 2019-12-04
Set boot order

zulcss:
In a multi-disk machine, boot order needs to be set, otherwise
the overcloud machines will fail to boot.

### DIFF
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -26,6 +26,12 @@
         template:
           src: "os-net-config-mappings.yaml.j2"
           dest: "/home/stack/os-net-config-mappings.yaml"
+      - name: Set boot order
+        shell: |
+          source ~/stackrc
+          openstack baremetal node list -c UUID -f value | \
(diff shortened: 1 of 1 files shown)


overcloud.yml
```diff
@@ -26,6 +26,12 @@
         template:
           src: "os-net-config-mappings.yaml.j2"
           dest: "/home/stack/os-net-config-mappings.yaml"
+      - name: Set boot order
+        shell: |
+          source ~/stackrc
+          openstack baremetal node list -c UUID -f value | \
+          xargs -I{} openstack baremetal node set {} --property root_device='{"name": "{{ root_disk }}"}'
```
smalleni, line 33: Not sure I follow, where are you getting the value of root_disk? I mean, how are you getting the disk name/WWN?

smalleni (changes requested):
Please address comments.

masco:

PR #142, opened 2019-12-21
optional undercloud as VM

masco:


main.yml
```diff
@@ -6,6 +6,16 @@
...
   tags:
     - always
+  when: virtual_uc != true
+
+- import_playbook: virtual_undercloud.yml
+  tags:
+    - always
+  when: virtual_uc == true
+
+- import_playbook: add_undercloud_to_inventory.yml
```
smalleni, line 16: If  virtual_uc is set to true, then this playbook is imported as part of setup_undercloud.yml. In that case, add_undercloud_to_inventory.yml is run a second time. Maybe have a when condition to this import?

tasks/load_instackenv.yml
```diff
@@ -24,25 +24,38 @@
...
+
+- name: set undercloud and overcloud info
+  block:
+    - name: set undercloud hostname
+      set_fact:
+        undercloud_hostname: "172.16.0.2"
+
+    - name: set overcloud_instackenv content
+      set_fact:
+          oc_instackenv_content: "{{ instackenv_content }}"
```
smalleni, line 55: So all the nodes in instackenv.json are assumed tob e overcloud nodes?

virtual_uc_network.yml
```diff
@@ -0,0 +1,41 @@
...
+# * br-ctlplane - used for provisioning
+# * management - external connectivity
+# * br-vlan - used for OSPd internal services (internal/tenent api)
+# * external - tenants external connectivity
+---
+networks:
+    net1:
+        name: br-ctlplane
+        forward: bridge
+        nic: p2p4
```
smalleni: Hardcoded?

tasks/load_instackenv.yml
```diff
@@ -24,25 +24,38 @@
...
+          oc_instackenv_content: |
+              {% set a=instackenv_content.pop('nodes') %}
+              {{ instackenv_content | combine({'nodes': a|difference([a[0]])}, recursive=True) }}
+  when: virtual_uc != true
+
+- name: set undercloud and overcloud info
+  block:
+    - name: set undercloud hostname
+      set_fact:
+        undercloud_hostname: "172.16.0.2"
```
smalleni: Maybe declare a var in group_vars/all.yml and use it here?

virtual_undercloud.yml
```diff
@@ -0,0 +1,104 @@
...
+  tasks:
+    - include_tasks: tasks/copykeys.yml
+      vars:
+        hostname: "localhost"
+        ssh_user: "root"
+
+    - name: clean previous vm
+      shell: |
+         source {{ infrared_dir }}/.venv/bin/activate
+         infrared virsh --host-address 127.0.0.1 --host-key /root/.ssh/id_rsa --topology-network virtual_uc_network --topology-nodes virtual_uc_topology:1 --host-memory-overcommit True --cleanup True
```
smalleni, line 22: So, looks like you are using localhost as the hypervisor. This will mean that I cannot run jetpack from my lapto etc right? I think what we can do is, in the case of baremetal we are excluding the first node as undercloud. We can do something similar in the case of virtualized undercloud and use the first node in isntackenv.json as the hypervisor. Thoughts?

smalleni (commented):
Looks like the assumption is to use the node from which jetpack is running as the hypervisor. I think it would be better to use the first node in instackenv.json as the hypervisor, as your current approach will fail, if for example I want to run jetpack from my  laptop.

Thoughts?

virtual_undercloud.yml
venkataanil: Agree. Same opinion.

virtual_undercloud.yml
```diff
@@ -0,0 +1,104 @@
...
+        ansible_ssh_private_key_file: "{{ ansible_ssh_key }}"
+        ansible_user: "cloud-user"
+        ansible_python_interpreter: "{{ python_interpreter }}"
+
+- hosts: undercloud
+  vars:
+    user: "stack"
+  gather_facts: yes
+  tasks:
+    - name: make sure line 'dns=none' is set in /etc/NetworkManager/NetworkManager.conf
```
venkataanil, line 55: This is nice work masco (as I know the context why you are playing with network manager).
Can you please add comment why you are doing this.

main.yml
venkataanil: Nice catch. @masco you forgot to remove add_undercloud_to_inventory.yml from setup_undercloud.yml.

undercloud.yml
```diff
@@ -2,28 +2,37 @@
...
+
+          - name: set undercloud_local_interface
+            set_fact:
+              undrecloud_local_interface: "{{ local_interface.stdout }}"
+        when: virtual_uc != true
 
       - name: set undercloud_local_interface
         set_fact:
-          undrecloud_local_interface: "{{ local_interface.stdout }}"
+          undrecloud_local_interface: "eth0"
```
venkataanil: Need to test with RHEL 8 image if it uses eth* naming convention. But for now it is fine. Please use variable.

external.yml
```diff
@@ -5,22 +5,31 @@
...
+            register: ext_interface
+          - name: set ext_iface
+            set_fact:
+              ext_iface: "{{ ext_interface.stdout }}"
+        when: virtual_uc != true
+
       - name: set ext_iface
         set_fact:
-          ext_iface: "{{ ext_interface.stdout }}"
+          ext_iface: "eth0"
```
venkataanil: variable (and also need to check with RHEL8) image.

virtual_uc_network.yml
venkataanil: use variable.

tasks/load_instackenv.yml
venkataanil: As discussed down, please add baremetal (in which undercloud vm is spawned ) as first node in instackenv.json

external.yml
masco: done

undercloud.yml
masco: done

virtual_undercloud.yml
masco: done

tasks/load_instackenv.yml
masco: done

virtual_uc_network.yml
masco: done

main.yml
masco: removed

templates/virtual_uc_topology.yml.j2
```diff
@@ -0,0 +1,25 @@
+name: undercloud
+prefix: null
+cpu: "8"
+cpumodel: "host-passthrough"
+memory: "24576"
+swap: "0"
+deploy_os: true
+os:
+    type: "linux"
+    variant: "rhel7"
```
smalleni, line 10: This needs to change, please parameterize this based on OSP version. rhel8 for OSP 15 and above.

smalleni (changes requested):
Please parameterize rhel version for UC VM

smalleni (approved):
OK to merge this with two changes for later revisit.

1. Add support to using first host of instackenv.json as the hypervisor
2. Support for RHEL8/OSP 16 (ansible jump host also RHEL 8)

PR #148, opened 2020-01-20
Fix ternary expression in prepare_nic_configs.yaml

smalleni:
The naming scheme is different for scale lab when compared to alias.
The ternary is now fixed to correctly get the machine type in case of alias.


masco:

Issue #136, opened 2019-12-03
overcloud tag failing as firstboot.yaml not found


venkataanil:
ansible-playbook main.yml -t overcloud failing with below error

TASK [find the template file or corresponding j2 template] *********************
^[[1;30mtask path: /root/.infrared/plugins/tripleo-overcloud/tasks/environment_templates.yml:6^[[0m
^[[0;34m<f16-h11-000-1029p.rdu2.scalelab.redhat.com> ESTABLISH SSH CONNECTION FOR USER: stack^[[0m
^[[0;34m<f16-h11-000-1029p.rdu2.scalelab.redhat.com> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o 'IdentityFile="/root/.ssh/id_rsa"' -o KbdInteractiveAuthentication
=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=stack -o ConnectTimeout=30 -o ControlPath=/root/.ansible/cp/5046f1bc61 f16-h11-000-1029p.rdu2
.scalelab.redhat.com '/bin/sh -c '"'"'echo ~stack && sleep 0'"'"''^[[0m
^[[0;34m<f16-h11-000-1029p.rdu2.scalelab.redhat.com> (0, '/home/stack\n', '')^[[0m
^[[0;34m<f16-h11-000-1029p.rdu2.scalelab.redhat.com> ESTABLISH SSH CONNECTION FOR USER: stack^[[0m
[...]

PR #155, opened 2020-02-13
Fix the location of extra templates

asyedham:
fixes the issue #136

### DIFF
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -22,10 +22,6 @@
       - name: set heat_template_version
         set_fact:
           heat_template_version: "{{ tht_version.stdout }}"
-      - name: generate os-net-config-mappings.yaml
-        template:
-          src: "os-net-config-mappings.yaml.j2"
-          dest: "/home/stack/os-net-config-mappings.yaml"
 
 - hosts: localhost
   gather_facts: yes
@@ -45,9 +41,18 @@
           - name: Copy files to undercloud if they do not exist
             copy:
               src: "{{ item.item }}"
-              dest: "/home/stack/"
+              dest: "/home/stack/extra_templates/"
             loop: "{{ env_files.results }}"
             when: not item.stat.exists
+            delegate_to: "{{ undercloud_hostname }}"
+            vars:
+              ansible_python_interpreter: "{{ python_interpreter }}"
+              ansible_user: "stack"
+
+          - name: generate os-net-config-mappings.yaml
+            template:
+              src: "os-net-config-mappings.yaml.j2"
+              dest: "/home/stack/extra_templates/os-net-config-mappings.yaml"
             delegate_to: "{{ undercloud_hostname }}"
             vars:
               ansible_python_interpreter: "{{ python_interpreter }}"
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -22,6 +22,10 @@
       - name: set heat_template_version
         set_fact:
           heat_template_version: "{{ tht_version.stdout }}"
+      - name: generate os-net-config-mappings.yaml
+        template:
+          src: "os-net-config-mappings.yaml.j2"
+          dest: "/home/stack/os-net-config-mappings.yaml"
 
 - hosts: localhost
   gather_facts: yes
@@ -44,15 +48,6 @@
               dest: "/home/stack/extra_templates/"
             loop: "{{ env_files.results }}"
             when: not item.stat.exists
-            delegate_to: "{{ undercloud_hostname }}"
-            vars:
-              ansible_python_interpreter: "{{ python_interpreter }}"
-              ansible_user: "stack"
-
-          - name: generate os-net-config-mappings.yaml
-            template:
-              src: "os-net-config-mappings.yaml.j2"
-              dest: "/home/stack/extra_templates/os-net-config-mappings.yaml"
             delegate_to: "{{ undercloud_hostname }}"
             vars:
               ansible_python_interpreter: "{{ python_interpreter }}"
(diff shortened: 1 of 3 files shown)


masco:
rerun tests

templates/overcloud-templates.yml.j2
```diff
@@ -1,13 +1,7 @@
 ---
 {% if extra_templates is defined and extra_templates %}
 tripleo_heat_templates:
-{% for template in  env_files.results %}
-{% if template.stat.exists == False %}
     - /home/stack/{{ template.item }}
```
masco: template.item is not defined

asyedham: Done

smalleni:
@asyedham Can you please explain what you are trying to do?

templates/overcloud-templates.yml.j2
```diff
@@ -2,11 +2,7 @@
 {% if extra_templates is defined and extra_templates %}
 tripleo_heat_templates:
 {% for template in  env_files.results %}
-{% if template.stat.exists == False %}
     - /home/stack/{{ template.item }}
-{% else %}
-    - {{ template.item }}
```
asyedham, line 8: @smalleni e.g if the template.stat.exists == true it is not able to find the firstboot.yaml file but if you specify it to be /home/stack/{{ template.item }} it resolves the issue and this fixes #136

masco: 1.what if the file doesn't exist there too?
2.what if the emplate.stat.exists == true is true?

i guess we need to swap the else part to true and provide the fully qualified name in the all.yaml.(ignore if it is wrong :-) )

asyedham: for the first time when we run an overcloud.yml the env_files.exists=false so we copy the firstboot.yaml to undercloud and then when we template out extra overcloud-templates [file](https://github.com/redhat-performance/jetpack/blob/master/templates/overcloud-templates.yml.j2) the path for tripleo_heat_templates is /home/stack//{{ template.item }}. But when we run it for the next time the firstboot.yml already exists on undercloud  as it was copied previously so the env_files.exists=true but due to the overcloud-templates.yml.j2 it tries to find the tripleo_heat_templates at {{ template.item }} . Also this path is passed in the overcloud_deploy.sh so the file should be located on undercloud i.e /home/stack/{{ template.item }}

smalleni: I will explain how this works and then you will understand why we need to take a different approach.

In this task at https://github.com/redhat-performance/jetpack/blob/master/overcloud.yml#L33 we check if the templates in the extra_templates in group_vars/all.yml exist on the undercloud.

The extra_templates variable list can have 
1. a relative path to an extra THT file which is in the jetpack directory on the host running jetpack (jump host) like firstboot.yaml that we currently have
2. an absolute path to a file like /usr/share/openstack-tripleo-heat-templates/environments/services/neutron-ovs-dpdk.yaml that exists on the undercloud node

Depending on whether or not the files in the extra_templates list exist on the undercloud, they are copied from the local directory of jetpack on the jumphost  to /home/stack/ on the undercloud if they do not exist. And then {{ infrared_dir }}/plugins/tripleo-overcloud/vars/overcloud/templates/extra.yml is templated out based on the location of file (absolute path if file exists or /home/stack if the file was copied from jetpack directory)

The issue described in https://github.com/redhat-performance/jetpack/issues/136 happens when
1. The THT file in the local jetpack directory on the ansible jump host has already been copied to /home/stack/<file.yaml> as it does not exist on the undercloud and then the playbook main.yml failed at some place. 
2. The user reruns the overcloud install by using ansible-playbook main.yaml --tags overcloud
3.  When the task at  https://github.com/redhat-performance/jetpack/blob/master/overcloud.yml#L35 runs, it runs as the stack user, so although it is searching for the firstboot.yaml relative path, it gets translated to /home/stack/firstboot.yml and finds the absolute file, so when it templates out the {{ infrared_dir }}/plugins/tripleo-overcloud/vars/overcloud/templates/extra.yml , it places "firstboot.yaml"  in the file, which is not the absolute path, so it does not resolve correctly. Bottom line is the extra.yml file we need absolute paths.

While you patch fixes the problem by always making path /home/stack/filename.yaml, it does not let user specify an existing THT file on the undercloud  in /usr/share/openstack-tripleo-heat-templates/.....

Does that make sense? We are losing the flexibility of specifying existing files on the undercloud. This patch to be re-worked and the problem might better be solved by changes in this template and/or tasks in the block at https://github.com/redhat-performance/jetpack/blob/master/overcloud.yml#L34

smalleni (changes requested):

templates/overcloud-templates.yml.j2
asyedham: Have updated the PR so that it fixes #136 and also provides the user the flexibility of specifying the existing files on undercloud

overcloud.yml
```diff
@@ -25,7 +25,7 @@
       - name: generate os-net-config-mappings.yaml
         template:
           src: "os-net-config-mappings.yaml.j2"
-          dest: "/home/stack/os-net-config-mappings.yaml"
+          dest: "/home/stack/extra_templates/os-net-config-mappings.yaml"
```
masco: you have to create the `extra_templates` folder before this task if the folder does not exist.

masco:
rerun tests

masco:
@asyedham the CI test is failing. please look that

asyedham:
@masco Fixed it

asyedham:
Closing this as we no longer use firstboot.yml playbook

PR #161, opened 2020-03-03
set master_doc

asyedham:
Need to set this to master_doc = 'index' to specify the document that contains the root toctree directive.


smalleni (approved):

smalleni:
@masco Why is the CI still running? Also we can ask CI to not run for any changes in docs, if that makes sense.

PR #165, opened 2020-03-06
Update foreman OS for RHEL 7

smalleni:
Scale lab foreman no longer supports just 'RHEL 7'. Minor version
needs to be specified.


venkataanil (approved):

venkataanil:
Thanks

Issue #174, opened 2020-04-01
virtual undercloud deployment fails if the lab_name is set to scale


asyedham:
When trying to install virtual_uc the deployment is failing if the lab_name is set to scale as it does not get machine_type https://github.com/redhatperformance/jetpack/blob/master/prepare_nic_configs.yml#L5

 As for virtual_uc the undercloud_hostname is "{{ undercloud_host }}"  i.e 172.16.0.2
https://github.com/redhat-performance/jetpack/blob/master/tasks/load_instackenv.yml#L51
```
TASK [set interfaces] ************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'item.key == machine_type and machine_type in scale.machine_types.supermicro' failed. The error was: error while evaluating conditional (item.key == machine_type and machine_type in scale.machine_types.supermicro): list object has no element 3\n\nThe error appears to be in '/root/jetpack/prepare_nic_configs.yml': line 21, column 11, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n      - block:\n        - name: set interfaces\n          ^ here\n"}

PLAY RECAP ***********************************************************************************************************************************************************************************
localhost                  : ok=44   changed=15   unreachable=0    failed=1    skipped=33   rescued=0    ignored=0   
undercloud                 : ok=10   changed=8    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0```

PR #175, opened 2020-04-03
Add support for virtual uc on first node

### LINKED ISSUE #173: Create a undercloud  VM on first host
Currently, undercloud VM created on the jump host i,e where you run your plays/jetpack. This should be moved to the first node of the list given in the instackenv.json.

asyedham:
fixes
#173 
#174

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -117,9 +117,10 @@ force_reprovision: false
 #do not enable this while running from your desktop
 virtual_uc: false
 undercloud_host: 172.16.0.2
+undercloud_hypervisor: redhat.com
 vm_external_interface: eth0
 undercloud_local_interface: eth0
-virtual_uc_ctlplane_interface: em1
+virtual_uc_ctlplane_interface: em2
 # RHEL 7 image url
 vm_image_url: https://url.corp.redhat.com/rhel-guest-image-7-6-210-x86-64-qcow2
 # RHEL 8 image url
--- a/tasks/load_instackenv.yml
+++ b/tasks/load_instackenv.yml
@@ -49,7 +49,16 @@
     - name: set undercloud hostname
       set_fact:
         undercloud_hostname: "{{ undercloud_host }}"
+
+    - name: first node in instackenv
+      set_fact:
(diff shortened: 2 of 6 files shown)


tasks/load_instackenv.yml
```diff
@@ -49,10 +49,33 @@
     - name: set undercloud hostname
       set_fact:
         undercloud_hostname: "{{ undercloud_host }}"
+        instack_firstnode: "{{ (instackenv_content.nodes[0].pm_addr | replace('mgmt-','') | replace('-drac', '')) }}"
```
masco, line 55: we need to give support for other labs also, since our CI is not running on these labs

asyedham: > we need to give support for other labs also, since our CI is not running on these labs

@masco can you please share the instackenv for those machines . I will add a task for it

asyedham:
@masco have added the option to specify the virtual undercloud. Could you please review

group_vars/all.yml
```diff
@@ -115,6 +115,7 @@ force_reprovision: false
 #This enables user to have a undercloud in VM
 #NOTE: now the vm is created in the ansible host, so
 #do not enable this while running from your desktop
+virtual_uc_hostname: redhat.com
```
masco: the above comment is for the below variable, so move this to one line below

masco: and make this variable optional, if users using scale or aliase lab, we no need to force the user to set this one.

asyedham:
@masco Could you please have look. If this needs any further changes as I'm testing it

virtual_undercloud.yml
```diff
@@ -1,6 +1,11 @@
 ---
 - hosts: undercloud_hypervisor
   tasks:
+    - name: Clean network interfaces
+      shell: |
+        /root/clean-interfaces.sh --nuke
+      changed_when: false
```
venkataanil, line 7: Will it throw error if clean-interfaces.sh not present on the hypervisor, for example, CI machine?

asyedham: Fixed it

PR #184, opened 2020-05-03
automate vm_image_url and virtual_uc_ctlpane_interface 

Shrutikabansal:
update the vm_image_url based on the osp version and virtual_uc_ctlpane_interface based on the machine type


masco:

Issue #186, opened 2020-05-04
ALIAS lab Dell 740xd network interface names don't exist


brainfunked:
```TASK [get ext_interface] *****************************************************************************************************************************************************
task path: /root/jetpack/external.yml:10
changed: [undercloud] => {"changed": true, "cmd": "for i in /sys/class/net/*\ndo\n  udevcontent=`udevadm info -p $i --query property`\n  if [[ $udevcontent =~ enp140s0f1 ]]\n  then\n    ext_i
nterface=`echo $i | cut -d '/' -f 5`\n    break\n  fi\ndone\necho $ext_interface\n", "delta": "0:00:00.058717", "end": "2020-05-04 14:20:04.191614", "rc": 0, "start": "2020-05-04 14:20:04.132
897", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
TASK [set ext_iface] *********************************************************************************************************************************************************
task path: /root/jetpack/external.yml:23
ok: [undercloud] => {"ansible_facts": {"ext_iface": ""}, "changed": false}
TASK [set ext_iface] *********************************************************************************************************************************************************
task path: /root/jetpack/external.yml:28
skipping: [undercloud] => {"changed": false, "skip_reason": "Conditional result was False"}
TASK [create vlan interface on external interface] ***************************************************************************************************************************
task path: /root/jetpack/external.yml:33
fatal: [undercloud]: FAILED! => {"changed": true, "cmd": "ip link add link  name .300 type vlan id 300\nip link set dev  up\nip link set dev .300 up\nip a a 172.17.5.1/24 brd 172.17.5.255 dev
 .300\n", "delta": "0:00:00.015098", "end": "2020-05-04 14:20:04.699105", "msg": "non-zero return code", "rc": 1, "start": "2020-05-04 14:20:04.684007", "stderr": "Cannot find device \"name\"
\nCannot find device \"up\"\nCannot find device \".300\"\nCannot find device \".300\"", "stderr_lines": ["Cannot find device \"name\"", "Cannot find device \"up\"", "Cannot find device \".300
\"", "Cannot find device \".300\""], "stdout": "", "stdout_lines": []}
...ignoring

```

This happens because the external interface is expected to match `enp140s0f1`, which doesn't exist. Neither does `enp140s0f0`. The correct naming for the interfaces should be `enp134s0f0` and `enp134s0f1`, which would match `ens7f0` and `ens7f1` respectively.

masco:
[...]

PR #187, opened 2020-05-04
Fixed alias lab dell 740xd interface naming

brainfunked:
Addressing #186


masco:

PR #196, opened 2020-05-18
Increase external network range

venkataanil:
We use the same CIDR(172.17.5.0/24) for floating ip in overcloud VMs.
But this won't be sufficient for scale testing. So use bigger
external network range.

Also set clean_debug to true by default, as most of the time
we will be deleting undercloud and redeploying.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -83,7 +83,7 @@ inspection_iprange: 192.168.24.110,192.168.24.250
 # undercloud to access overcloud resources
 external_gateway: 172.18.0.1/16
 external_network_vlan_id: 300
-clean_debug: True
+clean_debug: False
 #adding changes 
 external_net_cidr: 172.18.0.0/16
 external_allocation_pools_start: 172.18.0.50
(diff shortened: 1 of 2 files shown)


group_vars/all.yml
```diff
@@ -81,15 +81,14 @@ dhcp_end: 192.168.24.105
 inspection_iprange: 192.168.24.110,192.168.24.250
 # external network params for adding external network to
 # undercloud to access overcloud resources
-external_gateway: 172.17.5.1/24
-external_network_broadcast: 172.17.5.255
+external_gateway: 172.18.0.1/16
 external_network_vlan_id: 300
-clean_debug: False
+clean_debug: True
```
smalleni: Why this change?

venkataanil: This will clean the HDD when we delete the overcloud.
This will avoid booting the overcloud node from the unwanted disk when it has multiple disks.

smalleni: Ack

venkataanil: Lets have this as a separate change as it is impacting node intropsection.

PR #198, opened 2020-05-19
allow user to configure heat configs

masco:


group_vars/all.yml
```diff
@@ -80,6 +80,7 @@ scale:
 
 #extra_templates:
 
+#heat_configs:
```
asyedham, line 83: We can use the parameters_defaults in https://github.com/redhat-performance/jetpack/blob/master/group_vars/all.yml to add Tripleo Heat Templates configuration options

asyedham (changes requested):

PR #201, opened 2020-05-20
[WIP] Add support for ocp installation

asyedham:


asyedham:
Not yet tested. Also, we can divide it into sub playbooks if needed.

templates/ir_external_oc_subnet.yaml.j2
```diff
@@ -0,0 +1,19 @@
+---
+subnet:
+    cidr: "{{ external_net_cidr }}"
+    name: external_subnet
+    # Gateway for ext.net may differ from ExternalInterfaceDefaultRoute.
+    # As for ipv6 deployments we use still ipv4 gw, we need to specify
+    # it explicitly also here for ext. net. purpose.
+    gateway: "{{ external_gateway }}"
+    allocation_pool:
+        start: "{{ external_allocation_pools_start }}"
```
venkataanil: We can't use these values. They are dedicated for undercloud.
Thats the reason I was specifying different allocation pool here
https://docs.google.com/document/d/19fftnjqztw9R9sAZNvNTHXNbUBJ_UxhlsZk3y8NrFCI/edit#heading=h.54bp8w68ki2q

shift_stack.yml
```diff
@@ -0,0 +1,60 @@
+---
+- hosts: localhost
+  tasks:
+    - name: add masqerade rule
+      shell: |
+          sudo iptables -t nat -A POSTROUTING -o eno1 -j MASQUERADE
```
venkataanil: why is it eno1?

templates/ir_external_oc_subnet.yaml.j2
asyedham: Fixed it

PR #211, opened 2020-05-29
add nvme device support

venkataanil:
Openstack allows creating NVMe flavor and booting the VMs
with this flavor on compute nodes with NVMe PCI devices.
This change enhances jetpack to deploy director with
NVMe templates and creating NVMe flavor for the VM boot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,9 +83,58 @@ extra_templates:
   - example.yml
   - /usr/share/openstack-tripleo-heat-templates/environments/services/sahara.yaml
 parameter_defaults:
-  - NeutronOVSFirewallDriver: openvswitch`
+  NeutronServicePlugins: qos,ovn-router,trunk,segments
+  NeutronTypeDrivers: geneve,flat
+resource_registry:
+  - OS::TripleO::NodeUserData=/home/stack/firstboot-nvme.yaml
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -83,9 +83,58 @@ extra_templates:
...
+resource_registry:
+  - OS::TripleO::NodeUserData=/home/stack/firstboot-nvme.yaml
+  - OS::TripleO::Services::NeutronL3Agent: OS::Heat::None
+In the above case, jetpack copies firstboot-nvme.yaml from jetpack/files/firstboot-nvme.yaml to undercloud's /home/stack folder, if it doesn't exit in the undercloud at /home/stack/firstboot-nvme.yaml.
+
+
+*nvme* need to be configured to pass overcloud compute node's non-volatile memory express device directly to overcloud VM. Read [1] for details. ```flavor``` is used to create openstack flavor after the deployment in Jetpack's post module. We can get the nvme device details with the below script
+cat ./get_nvme_address.sh
```
masco, line 113: this line can be removed and the below script can be shown in a code block or we can have a helpers dir(or any other good name also fine) and add this kind of scripts.

venkataanil: agree

PR #216, opened 2020-06-05
support lab public network

venkataanil:
We can request scale lab team for external public network and use
this for accessing overcloud VMs from outside LAB environment.

1) user has to set public_external_interface for jetpack to use it
   for OSP director deployment
2) user need to define external_interface, which specifies the
 interface on which lab team has enabled public network
3) define CIDR and other parameters according to README.md
4) This interface is a access port and user shouldn't define
   VLANs on it
5) jetpack will disable rp_filter on all nodes. This is needed to
   allow traffic through this public interface.
6) jetpack is enhanced to create overcloud external network after
   overcloud deployment. When public network is enabled, jetpack
   will create the network as neutron flat provider network.

### DIFF
--- a/external.yml
+++ b/external.yml
@@ -25,10 +25,28 @@
               ext_iface: "{{ ext_interface.stdout }}"
         when: virtual_uc != true
 
+      - name: fail when both public_external_interface and external_network_vlan_id defined
+        fail:
+          msg: Both public_external_interface and external_network_vlan_id defined
+        when: public_external_interface is defined and external_network_vlan_id is defined
+
       - name: set ext_iface
         set_fact:
           ext_iface: "{{ vm_external_interface }}"
         when: virtual_uc == true
+
+      # "Reverse path" filter prevents packets from coming in or going out on
+      #  an unexpected interface to help prevent routing loops (check with lab team).
+      # The default route on undercloud was out the primary foreman interface,
+      # so when undercloud gets packets coming in on the 4th NIC, it sends the traffic back
+      # out the foreman interface. Turning off this configuration enables to use
+      # public_external_interface i.e 4th interface
+      - name: disable rp_filter
+        shell: |
+            for f in $(find /proc/sys/net/ipv4 -name rp_filter) ; do echo 0 > $f ; done
+        become: true
+        ignore_errors: true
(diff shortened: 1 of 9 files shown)


external.yml
```diff
@@ -25,30 +25,58 @@
...
           ext_iface: "{{ vm_external_interface }}"
         when: virtual_uc == true
 
+      # "Reverse path" filter prevents packets from coming in or going out on
+      #  an unexpected interface to help prevent routing loops (check with lab team).
+      # The default route on undercloud was out the primary foreman interface,
+      # so when undercloud gets packets coming in on the 4th NIC, it sends the traffic back
+      # out the foreman interface. Turning off this configuration enables to use
+      # public_external_interface i.e 4th interface
+      - name: enable rp_filter
```
smalleni: This task is disabling rp filter not enabling.

group_vars/all.yml
```diff
@@ -70,6 +70,16 @@ scale:
...
 
+# Scale lab can provide us public external network on 4th interface
+# We can use that to access overcloud VM's from outside undercloud.
+# To use this feature set 'public_external_interface: true' and
+# assign 4th nic to external_interface to 4th nic and
+# define external_XXXXXX parameters wth CIDR provided by lab team
+# Don't define external_network_vlan_id as lab team might have
+# created their own vlan for this external network implementation.
+#public_external_interface: false
+#external_interface: ens2f3
```
smalleni, line 84: You should make this clear, if its a rhel7 interface or rhel 8 interface.

smalleni (changes requested):

group_vars/all.yml
venkataanil: I have added that in README.md
https://github.com/redhat-performance/jetpack/pull/216/files#diff-04c6e90faac2675aa89e2176d2eec7d8R147

external.yml
venkataanil: agree

group_vars/all.yml
smalleni: Yes, but I don't think it mentiones anything about rhel8/rhel7 naming convention. Atleast, you might want to add that in the README?

PR #219, opened 2020-06-17
Honor {{ ansible_ssh_key }} var when copying keys

smalleni:

### DIFF
--- a/tasks/copykeys.yml
+++ b/tasks/copykeys.yml
@@ -1,7 +1,7 @@
 ---
 - name: copy jump host key
   shell: |
-    echo '{{ ansible_ssh_pass }}' | sshpass ssh-copy-id  -o 'StrictHostKeyChecking no' -f {{ ssh_user }}@{{ hostname }}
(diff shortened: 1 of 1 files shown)


tasks/copykeys.yml
```diff
@@ -1,15 +1,15 @@
 ---
 - name: copy jump host key
   shell: |
-    echo '{{ ansible_ssh_pass }}' | sshpass ssh-copy-id  -o 'StrictHostKeyChecking no' -f {{ ssh_user }}@{{ hostname }}
+    echo '{{ ansible_ssh_pass }}' | sshpass ssh-copy-id -i {{ ansible_ssh_key }} -o 'StrictHostKeyChecking no' -f {{ ssh_user }}@{{ hostname }}
```
venkataanil, line 4: Why is this needed?

smalleni: Otherwise it always copies the user’s default keys and does not honor if you give a custom ansible_ssh_key var. right now this always works since it just copied the default keys and we don’t give it a custom path for ansible_ssh_key

venkataanil: Thanks

PR #226, opened 2020-07-08
Containerize jetpack

asyedham:


masco:
can you start the jetpack with debug mode and capture the log in some file, so that if execution fails user can look the logs and find the reason for failure.

asyedham:
> can you start the jetpack with debug mode and capture the log in some file, so that if execution fails user can look the logs and find the reason for failure.

I can add an option to run the playbook in verbose mode so that the user can view the reason for the failure.

PR #245, opened 2020-09-11
use rhel7_8 for osp13 UC

masco:


masco:
rerun tests

masco:
rerun tests

PR #250, opened 2020-09-23
install os asynchronously

venkataanil:
When Jetpack has to install OS on several machines, this will speed
up only when run in parallel. Scale cloud simulation requires
CentOS  on all the lab machines, so we need to install
asynchronously.

### DIFF
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -10,11 +10,25 @@
         16: 8.1
         16.1: 8.2
   tasks:
-    - include_tasks: tasks/install_os.yml
+    - name: set async_install to empty list
+      set_fact:
+        async_install: []
+
+    - name: install os on undercloud node
       vars:
         needed_os: "RHEL {{ osp_rhel_mapping[osp_release|float] }}"
         hypervisor_host: "{{ undercloud_hostname }}"
         hypervisor_password: "{{ ansible_ssh_pass }}"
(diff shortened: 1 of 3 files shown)


setup_undercloud.yml
```diff
@@ -10,11 +10,25 @@
         16: 8.1
         16.1: 8.2
   tasks:
-    - include_tasks: tasks/install_os.yml
+    - name: set async_install to empty list
```
smalleni, line 13: This is not needed if you use
async_install: "{{ async_install|default([], true) + [hyp_reachable.ansible_job_id] }}" in tasks/install_os.yaml

venkataanil: When we run multi hypervisor, the order of playbooks is like below
1. scale_compute_vms will install os on the hypervisors, which will set async_install like below
    async_install: ['hyp3', 'hyp4']
2. After scale_compute_vms playbook is finished, os is already installed on the hypervisors, async_install still will have these entries though not needed in later playbooks
2. now when setup_undercloud called to install OS on undercloud machine, if I am not emptying here, my async_install will become
    async_install: ['hyp3', 'hyp4', 'undercloud']
I will be ending up checking status for all the above nodes again, which is not needed.

smalleni: Got it. I didn't follow the scale_compute_vms flow. Thanks for clarifying.

PR #266, opened 2020-10-20
Update OS to RHEL 7.9 for OSP13

asyedham:


asyedham:
rerun

setup_undercloud.yml
```diff
@@ -4,7 +4,7 @@
       chassis_password:  "{{ instackenv_content.nodes[0].pm_password }}"
       osp_rhel_mapping:
         10: 7.7
-        13: 7.7
+        13: 7.9
```
masco, line 7: we will keep this as 7.8 and update it to 7.9 for OSP13. it will be very simple and straightforward. no need to revert to 7.8 from 7.9 is what I am thinking. may be we can add some comments about, why are we doing this.

asyedham: Yeah, but user can get osp_rhel_mapping  at a place with exact osp_release and required rhel_version. At the first sight, it can cause a misunderstanding that we are using RHEL 7.8 for OSP13.

PR #279, opened 2020-10-28
Enable node cleaning by default for OSP > 13

smalleni:

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -118,7 +118,7 @@ inspection_iprange: 192.168.24.110,192.168.24.250
 # undercloud to access overcloud resources
 external_gateway: 172.17.5.1/24
 external_network_vlan_id: 300
-clean_debug: False
+clean_nodes: True
 #adding changes 
 external_net_cidr: 172.17.5.0/24
 external_allocation_pools_start: 172.17.5.50
--- a/templates/undercloud.conf.j2
+++ b/templates/undercloud.conf.j2
@@ -2,7 +2,7 @@
 local_interface = {{ undercloud_local_interface }}
 local_ip = {{ gateway }}/{{ subnet_mask }}
 undercloud_ntp_servers = clock.redhat.com,clock2.redhat.com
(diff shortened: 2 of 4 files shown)


smalleni:
@asyedham @venkataanil PTAL

akrzos (approved):
LGTM

templates/undercloud.conf.j2
```diff
@@ -11,6 +11,7 @@ container_insecure_registries={{ registry_mirror }},{{ insecure_registries }}
 undercloud_timezone = UTC
 undercloud_public_host = {{ undercloud_public_host }}
 undercloud_admin_host = {{ undercloud_admin_host }}
+clean_nodes = true
```
asyedham: Instead, we can set the clean_debug to true by default in [group_vars/all.yml](https://github.com/redhat-performance/jetpack/blob/master/group_vars/all.yml#L121) and disable the node cleaning if needed by setting it to false.
This way there is no need for setting clean_nodes to true again. Thoughts?

asyedham (changes requested):

templates/undercloud.conf.j2
smalleni: Oh sweet, I didn’t even check that we already have that variable. Yeh let me just change the default there. Good point.

group_vars/all.yml
```diff
@@ -118,7 +118,7 @@ inspection_iprange: 192.168.24.110,192.168.24.250
 # undercloud to access overcloud resources
 external_gateway: 172.17.5.1/24
 external_network_vlan_id: 300
-clean_debug: False
+clean_nodes: True
```
masco, line 121: can you change this on CI vars also so you will get a green mark from CI ;-)

smalleni: Sure, wasn't sure if you wanted node cleaning in CI :)

smalleni:
Done with requested changes.

masco (approved):
waiting for CI result

PR #283, opened 2020-10-29
Override max_concurrent_builds set by infrared

### LINKED ISSUE #272: Bump max concurrent builds to speed up overcloud deployment
Looks like [infrared](https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/deploy.yml#L62) specifically reduced the number of nodes that can concurrently be built to 5, when we could speed this process up for consumers of the scale lab with larger undercloud machines.  We should set this back to 10 or work with infrared to allow us to have an override for the scale lab.

smalleni:
Infrared changes the max_concurrent_builds from 10 to 5.
This commit helps set that back to 10 for faster deploys.

Fixes: #272

### DIFF
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -28,6 +28,28 @@
           dest: "/home/stack/firstboot-nvme.yaml"
         when: passthrough_nvme is defined or mount_nvme is defined
 
+      - block:
+          - name: adjust max number of concurrent builds to default (override infrared)
+            ini_file:
+                path: /var/lib/config-data/puppet-generated/nova/etc/nova/nova.conf
+                section: DEFAULT
+                option: max_concurrent_builds
+                value: 10
+            become: yes
+
+          - name:  get nova services
+            shell: systemctl list-units | grep 'nova.*.service' |  awk {'print$1'}
+            register: nova_services
+            become: yes
+
+          - name: restart nova services
+            systemd:
+              name: "{{ item }}"
+              state: restarted
+            with_items: "{{ nova_services.stdout_lines }}"
+            become: yes
+        when: osp_release >= 12
+
 - hosts: localhost
   gather_facts: yes
   tasks:
@@ -84,21 +106,21 @@
             NovaSchedulerAvailableFilters: ['nova.scheduler.filters.all_filters'],
             NovaSchedulerMaxAttempts: "{{ compute_count }}",
             NovaPCIPassthrough: [{
-                vendor_id: "{{ nvme.vendor_id }}",
-                product_id: "{{ nvme.product_id }}",
-                address: "{{ nvme.address }}" }],
+                vendor_id: "{{ passthrough_nvme.vendor_id }}",
+                product_id: "{{ passthrough_nvme.product_id }}",
+                address: "{{ passthrough_nvme.address }}" }],
             ControllerExtraConfig: {
                 "nova::pci::aliases": [{
                     name: 'nvme',
-                    vendor_id: "{{ nvme.vendor_id }}",
-                    product_id: "{{ nvme.product_id }}",
+                    vendor_id: "{{ passthrough_nvme.vendor_id }}",
+                    product_id: "{{ passthrough_nvme.product_id }}",
                     device_type: 'type-PCI'}]
             },
             ComputeExtraConfig: {
                 "nova::pci::aliases": [{
                     name: 'nvme',
-                    vendor_id: "{{ nvme.vendor_id }}",
-                    product_id: "{{ nvme.product_id }}",
+                    vendor_id: "{{ passthrough_nvme.vendor_id }}",
+                    product_id: "{{ passthrough_nvme.product_id }}",
                     device_type: 'type-PCI'}]
              },
              ComputeKernelArgs: "intel_iommu=on iommu=pt",
--- a/post.yml
+++ b/post.yml
@@ -6,7 +6,7 @@
     - name: create nvme flavor
       shell: |
         source /home/stack/overcloudrc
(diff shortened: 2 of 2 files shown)


smalleni:
@asyedham The only question I have is if the nova services on undercloud in OSP 13 also follow the naming covention of "tripleo_nova_*" . Could you please verify on an OSP 13 build?

asyedham:
> @asyedham The only question I have is if the nova services on undercloud in OSP 13 also follow the naming covention of "tripleo_nova_*" . Could you please verify on an OSP 13 build?

@smalleni Nova services are not following the naming convention of "tripleo_nova_*".  You can use "openstack-nova-*"

overcloud.yml
```diff
@@ -27,6 +27,19 @@
...
+          - name: adjust max number of concurrent builds to default (override infrared)
+            ini_file:
+                path: /var/lib/config-data/puppet-generated/nova/etc/nova/nova.conf
+                section: DEFAULT
+                option: max_concurrent_builds
+                value: 10
+            become: yes
+
+          - name: restart nova services
+            command: "systemctl restart tripleo_nova_*"
```
asyedham: use "systemctl restart openstack-nova-*"

asyedham (changes requested):

overcloud.yml
```diff
@@ -27,6 +27,19 @@
           src: "firstboot-nvme.yaml.j2"
           dest: "/home/stack/firstboot-nvme.yaml"
         when: nvme is defined
+      - block:
+          - name: adjust max number of concurrent builds to default (override infrared)
+            ini_file:
+                path: /var/lib/config-data/puppet-generated/nova/etc/nova/nova.conf
+                section: DEFAULT
+                option: max_concurrent_builds
+                value: 10
```
asyedham, line 37: Instead of hardcoding, we can parameterize this value.

asyedham (changes requested):

overcloud.yml
akrzos: On my 16.1 Undercloud the services for nova are tripleo_nova_*

```
[stack@undercloud ~]$ sudo systemctl list-units | grep nova
  tripleo_nova_api.service                                                                                                             loaded active running   nova_api container                                                                                                          
  tripleo_nova_api_cron.service                                                                                                        loaded active running   nova_api_cron container                                                                                                     
  tripleo_nova_compute.service                                                                                                         loaded active running   nova_compute container                                                                                                      
  tripleo_nova_conductor.service                                                                                                       loaded active running   nova_conductor container                                                                                                    
  tripleo_nova_scheduler.service                                                                                                       loaded active running   nova_scheduler container                                                                                                    
  tripleo_nova_api_cron_healthcheck.timer                                                                                              loaded active waiting   nova_api_cron container healthcheck                                                                                         
  tripleo_nova_api_healthcheck.timer                                                                                                   loaded active waiting   nova_api container healthcheck                                                                                              
  tripleo_nova_compute_healthcheck.timer                                                                                               loaded active waiting   nova_compute container healthcheck                                                                                          
  tripleo_nova_conductor_healthcheck.timer                                                                                             loaded active waiting   nova_conductor container healthcheck                                                                                        
  tripleo_nova_scheduler_healthcheck.timer                                                                                             loaded active waiting   nova_scheduler container healthcheck                                                                                        
```

overcloud.yml
smalleni: Probably not, I don't see us changing this value very often. If we need to let's do it at that point.

overcloud.yml
smalleni: The updated logic should work for both cases.

overcloud.yml
```diff
@@ -27,6 +27,27 @@
...
+          - name:  get nova services
+            command: systemctl list-units | grep 'nova.*.service'
+            register: nova_services
+            become: yes
+
+          - name: restart nova services
+            systemd:
+              name: "{{ item }}"
+              state: restarted
+            with_items: "{{ nova_services.stdout_lines }}"
```
masco: trailing whitespace causing lint failure

masco:
rerun tests

overcloud.yml
```diff
@@ -27,6 +27,27 @@
...
+          - name: adjust max number of concurrent builds to default (override infrared)
+            ini_file:
+                path: /var/lib/config-data/puppet-generated/nova/etc/nova/nova.conf
+                section: DEFAULT
+                option: max_concurrent_builds
+                value: 10
+            become: yes
+
+          - name:  get nova services
+            command: systemctl list-units | grep 'nova.*.service'
```
asyedham: you can use awk in the above command as it is currently not just fetching the name of the nova services instead it registering 
```
[stack@undercloud ~]$ systemctl list-units --all | grep 'nova.*.service'
  openstack-nova-api.service                                                                                     loaded    active   running   OpenStack Nova API Server
  openstack-nova-compute.service                                                                                 loaded    active   running   OpenStack Nova Compute Server
  openstack-nova-conductor.service                                                                               loaded    active   running   OpenStack Nova Conductor Server
  openstack-nova-scheduler.service                                                                               loaded    active   running   OpenStack Nova Scheduler Server
```
So the OSP13 CI job is failing

asyedham (changes requested):

overcloud.yml
smalleni: My bad, I missed the awk...

smalleni:
@masco @venkataanil @asyedham Figured out the issue. I was using command module isntead of shell. Now that is fixed, the failure in CI is in rally setup. Looks like you need to fix that, but this commit can be merged.

PR #286, opened 2020-10-31
use nvme for nova ephemeral disk

venkataanil:
This will mount nvme device on compute node and will use that as
nova ephemeral storage so that instances are booted on nvme device.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -168,9 +168,14 @@ shift_stack: false
 #installs browbeat, if this variable is not defined or set true
 #skips browbeat installation, only if the variable set false
 browbeat: true
+
+# define mount_nvme to mount /var/lib/nova on nvme device
+# to use nvme storage for nova instances as ephemeral disk
+#mount_nvme: true
+
 # enable nvme parameters
 # README has details about how to get these values
-#nvme:
+#passthrough_nvme:
 #    vendor_id: '144d'
 #    product_id: 'a804'
 #    address: '01:00.0'
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -26,7 +26,7 @@
         template:
           src: "firstboot-nvme.yaml.j2"
           dest: "/home/stack/firstboot-nvme.yaml"
-        when: nvme is defined
+        when: passthrough_nvme is defined or mount_nvme is defined
 
 - hosts: localhost
   gather_facts: yes
@@ -105,14 +105,14 @@
           }
         set_fact:
           parameter_defaults: "{{ parameter_defaults|default({}) | combine(nvme_params) }}"
-        when: nvme is defined
+        when: passthrough_nvme is defined
 
       - name: set resource_registry nvme parameters
         vars:
           res_reg: "{{ resource_registry|default([]) }}"
         set_fact:
           resource_registry: "{{ res_reg + ['OS::TripleO::NodeUserData=/home/stack/firstboot-nvme.yaml'] }}"
-        when: nvme is defined
+        when: passthrough_nvme is defined or mount_nvme is defined
 
       - name: add tripleo_heat_templates to template
         set_fact:
--- a/templates/firstboot-nvme.yaml.j2
+++ b/templates/firstboot-nvme.yaml.j2
@@ -31,9 +31,18 @@ resources:
             set -x
             FORMAT='compute'
             if [[ $(hostname) == *$FORMAT* ]] ; then
-              sed "s/^\(GRUB_CMDLINE_LINUX=\".*\)\"/\1 $KERNEL_ARGS\"/g" -i /etc/default/grub ;
-              grub2-mkconfig -o /etc/grub2.cfg
-              reboot
+              {% if mount_nvme is not defined and passthrough_nvme is defined %} 
+                sed "s/^\(GRUB_CMDLINE_LINUX=\".*\)\"/\1 $KERNEL_ARGS\"/g" -i /etc/default/grub ;
+                grub2-mkconfig -o /etc/grub2.cfg
+                reboot
+              {% endif %}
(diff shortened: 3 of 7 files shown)


templates/firstboot-nvme.yaml.j2
```diff
@@ -33,6 +33,11 @@ resources:
             if [[ $(hostname) == *$FORMAT* ]] ; then
               sed "s/^\(GRUB_CMDLINE_LINUX=\".*\)\"/\1 $KERNEL_ARGS\"/g" -i /etc/default/grub ;
               grub2-mkconfig -o /etc/grub2.cfg
+              mkdir -p /var/lib/nova
+              mkfs -t xfs -f /dev/nvme0n1
+              mount /dev/nvme0n1 /var/lib/nova
+              chmod a+w /var/lib/nova
+              echo UUID=`sudo blkid -s UUID -o value /dev/nvme0n1` /var/lib/nova xfs discard,defaults,nofail 0 2 | sudo tee -a /etc/fstab
```
smalleni: Do you plan to hardcode /dev/nvme0n1 here?

group_vars/all.yml
```diff
@@ -167,6 +167,11 @@ shift_stack: false
 #installs browbeat, if this variable is not defined or set true
 #skips browbeat installation, only if the variable set false
 browbeat: true
+
+# define mount_nvme to mount /var/lib/nova on nvme device
+# to use nvme storage for nova instances as ephemeral disk
+#mount_nvme: true
```
smalleni, line 174: I recommend mount_nvme to be a dictionary with one of the keys being the path of the nvme device (/dev/nvmen01..). Similarly for the passthrough stuff, we should change the name from `nvme` to something like `passthrough_nvme` to distinguish

templates/firstboot-nvme.yaml.j2
smalleni: Do you plan to keep the nvme path hardcoded? Would be good to parameterize inside of the `mount_nvme` dictionary.

overcloud.yml
```diff
@@ -26,7 +26,7 @@
         template:
           src: "firstboot-nvme.yaml.j2"
           dest: "/home/stack/firstboot-nvme.yaml"
-        when: nvme is defined
+        when: nvme is defined or mount_nvme is defined
```
smalleni: The template also have some fields like ComputeKernelArgs that are only needed for the passthrough case, so we should have a condition within the template.

templates/firstboot-nvme.yaml.j2
```diff
@@ -33,6 +33,11 @@ resources:
             if [[ $(hostname) == *$FORMAT* ]] ; then
               sed "s/^\(GRUB_CMDLINE_LINUX=\".*\)\"/\1 $KERNEL_ARGS\"/g" -i /etc/default/grub ;
               grub2-mkconfig -o /etc/grub2.cfg
+              mkdir -p /var/lib/nova
+              mkfs -t xfs -f /dev/nvme0n1
+              mount /dev/nvme0n1 /var/lib/nova
+              chmod a+w /var/lib/nova
```
smalleni: Also you can't do this mounting if you want to passthrough nvme as the device won't be available on the hypervisor. Looks like you need a condition within the template to do this only for the case where you mount nvme.

smalleni (changes requested):

templates/firstboot-nvme.yaml.j2
smalleni: Is the file system xfs even on OSP 13 ? I rememberthat the default is different between OSP 13 and OSP 16, you might want to check.

templates/firstboot-nvme.yaml.j2
venkataanil: nvme0n1 naming is used for nvme device. We can use only one nvme device for nova instances.
This hard coding will be a problem only when we have multiple nvme devices and other nvme devices are of bigger size than the first nvme device.

templates/firstboot-nvme.yaml.j2
venkataanil: Agree. As we know, NVME passthrough is introduced for OCP. As OCP can't use it (OCP IPI limitation), thought of removing NVME passthrough in later patches.

venkataanil:
Sai, I agree for all your suggestions. I will push a new patchset.
Masco, you can still test with this first patchset.

templates/firstboot-nvme.yaml.j2
venkataanil: > Is the file system xfs even on OSP 13 ? I rememberthat the default is different between OSP 13 and OSP 16, you might want to check.

Both OSP13 and OSP16.1 are using xfs file systems.

masco:
i am deploying with nvme disabled

links84:
I was hoping if jetpack finds more nvme disk, it should create and lvm with all nvme disk and mount it.

venkataanil:
> I was hoping if jetpack finds more nvme disk, it should create and lvm with all nvme disk and mount it.

Can you please elaborate?

links84:
> > I was hoping if jetpack finds more nvme disk, it should create and lvm with all nvme disk and mount it.
> 
> Can you please elaborate?

We have nodes with 4 nvme disks too. In future if we have higher storage requirements per compute, we can use such nodes and make sure all 4 nvme disks are added to lvm and mounted .
We can use stripe = numer of disk to improve performance.
lvcreate .. -i <number of disks>  <VG name>

group_vars/all.yml
venkataanil: we can think about using as dict later when we get other requirements

masco: @venkataanil can you add these variables in CI files too.

masco:
rerun tests

PR #297, opened 2020-11-12
restructure the tags

asyedham:
This PR helps to redefine the tags so that only the required facts are generated while using tags.
Example:
ansible-playbook main.yaml -t undercloud
If the playbook main.yml is already run once and then user wants to use tag undercloud, it triggers only some specific tasks to deploy undercloud. It similarly works for the other tags as well.

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -14,7 +14,14 @@
         when: sshkey.stat.exists == False
 
       - name: load instackenv file
-        include_tasks: tasks/load_instackenv.yml
+        import_tasks: tasks/load_instackenv.yml
+        tags:
+          - undercloud
+          - introspect
+          - tag
+          - composable
+          - external
+          - overcloud
 
       - name: setup infrared
         include_tasks: tasks/setup_infrared.yml
@@ -53,6 +60,7 @@
         vars:
           lab_vars: "{{ (lab_name == 'scale') | ternary(scale, alias) }}"
         when: lab_name in ['scale', 'alias']
+        tags: undercloud, introspect, tag, composable, external, overcloud
 
       - block:
           - name: set machines_types of overcloud nodes
@@ -70,6 +78,7 @@
         vars:
           lab_vars: "{{ (lab_name == 'scale') | ternary(scale, alias) }}"
         when: lab_name in ['scale', 'alias'] and virtual_uc != true
+        tags: undercloud, introspect, tag, composable, external, overcloud
 
       # use of a local time server is essential or at least a good idea for Ceph, Kerberos, Redis,
       # maybe Kubernetes, and other components sensitive to server time skew
--- a/cleanup.yml
+++ b/cleanup.yml
@@ -4,26 +4,32 @@
         file:
             path: "{{ infrared_dir }}"
             state: absent
+
       - name: remove {{ infrared_workspaces_dir }}
         file:
             path: "{{ infrared_workspaces_dir }}"
             state: absent
+
       - name: remove {{ instackenv_file }}
         file:
             path: "{{ instackenv_file }}"
             state: absent
+
       - name: remove overcloud_instackenv.json
         file:
             path: "~/overcloud_instackenv.json"
             state: absent
+
       - name: remove nic-configs virt
         file:
             path: "{{ ansible_user_dir }}/virt"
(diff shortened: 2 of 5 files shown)


venkataanil (commented):
With this change you will be ending up creating new workspace instead of using the existing work space.

My suggestion is to not to call this task file at all when both ~/infrared and ~/.infrared exists and instead use the existing workspace

- include_tasks: tasks/setup_infrared.yml
  when: infrared_dir.stat.exists ==False and infrared_workspace.stat.exists == False

asyedham:
> With this change you will be ending up creating new workspace instead of using the existing work space.
> 
> My suggestion is to not to call this task file at all when both ~/infrared and ~/.infrared exists and instead use the existing workspace
> 
> * include_tasks: tasks/setup_infrared.yml
>   when: infrared_dir.stat.exists ==False and infrared_workspace.stat.exists == False

Oh, got it. I will work on it. Thanks

venkataanil (commented):
Also I think we should avoid calling setup_undercloud.yml and add_undercloud_to_inventory.yml playbooks when we are using existing inventory. Think about redefining playbooks which uses always tag,
when we need to reuse existing inventory and run either undercloud, introspect,tag,overcloud,post tags - don't need to
1. install packages
2. setup infrared
3. generate keys
4. setup undercloud
5. add undercloud to inventory
We only need to regenerate facts i.e vendors, machine_types, undercloud_hostname, oc_instackenv_content, ifaces etc..

asyedham:
> Also I think we should avoid calling setup_undercloud.yml and add_undercloud_to_inventory.yml playbooks when we are using existing inventory. Think about redefining playbooks which uses always tag,
> when we need to reuse existing inventory and run either undercloud, introspect,tag,overcloud,post tags - don't need to
> 
> 1. install packages
> 2. setup infrared
> 3. generate keys
> 4. setup undercloud
> 5. add undercloud to inventory
>    We only need to regenerate facts i.e vendors, machine_types, undercloud_hostname, oc_instackenv_content, ifaces etc..

I have redefined the tags to re-use the existing inventory and only regenerate the required facts while running with the above changes.  Please have a look

bootstrap.yml
```diff
@@ -16,8 +16,18 @@
...
+        stat:
+          path: "{{ item }}"
+        register: infrared_dirs
+        loop:
+          - "{{ infrared_dir }}"
+          - "{{ infrared_workspaces_dir }}"
+
       - name: setup infrared
         include_tasks: tasks/setup_infrared.yml
+        when: item.stat.exists == False
```
venkataanil: So if both ~/infrared and ~/.infrared doesn't exist, we are calling setup_infrared.yml twice, right?

bootstrap.yml
```diff
@@ -16,8 +16,18 @@
...
+        with_items: "{{ infrared_dirs.results }}"
 
       - name: add localhost to inventory file
```
venkataanil: When you already have inventory in ~/.infrared/.workspaces/active/hosts, why you want to prepare again?

venkataanil (commented):
Can you please also add more details in commit message i.e

What you are trying to do and how you are achieving that? Its hard to just go through the code and understand what you are trying to do.

venkataanil (commented):
Can you move this as a separate PR?

asyedham:
> Can you move this as a separate PR?

Done, I have created new PR for checking the infrared dirs
https://github.com/redhat-performance/jetpack/pull/307

And I have updated the comment at top with brief details of the PR

links84 (approved):

cleanup.yml
```diff
@@ -4,26 +4,32 @@
         file:
             path: "{{ infrared_dir }}"
             state: absent
-      - name: remove {{ infrared_workspaces_dir }}
+
+      - name: remove .infrared dir
         file:
-            path: "{{ infrared_workspaces_dir }}"
+            path: "{{ ansible_user_dir }}/.infrared"
```
masco: I am seeing these changes in another PR also. i think it got missed while separating the works.

asyedham: Done

asyedham:
@masco I have rebased it. Please have a look

PR #307, opened 2020-11-20
check for infrared dirs

asyedham:
This PR checks if the infrared dirs already exist on the system
If it exists skips tasks/setup_infrared.yml
If not triggers tasks/setup_infrared.yml

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -16,8 +16,19 @@
       - name: load instackenv file
         include_tasks: tasks/load_instackenv.yml
 
+      - name: check if the infrared dir exists
+        stat:
+          path: "{{ infrared_dir }}"
+        register: infrared_directory
+
+      - name: check if the infrared workspace dir exist
+        stat:
+          path: "{{ infrared_workspaces_dir }}"
+        register: infrared_workspaces_directory
+
       - name: setup infrared
         include_tasks: tasks/setup_infrared.yml
+        when: infrared_directory.stat.exists == False or infrared_workspaces_directory.stat.exists == False
 
       - name: add localhost to inventory file
         add_host:
--- a/cleanup.yml
+++ b/cleanup.yml
@@ -4,26 +4,31 @@
         file:
             path: "{{ infrared_dir }}"
             state: absent
-      - name: remove {{ infrared_workspaces_dir }}
+      - name: remove .infrared dir
         file:
-            path: "{{ infrared_workspaces_dir }}"
+            path: "{{ ansible_user_dir }}/.infrared"
             state: absent
+
       - name: remove {{ instackenv_file }}
         file:
             path: "{{ instackenv_file }}"
             state: absent
+
       - name: remove overcloud_instackenv.json
         file:
             path: "~/overcloud_instackenv.json"
             state: absent
+
       - name: remove nic-configs virt
         file:
             path: "{{ ansible_user_dir }}/virt"
             state: absent
+
       - name: remove nic-configs virt_4nics
         file:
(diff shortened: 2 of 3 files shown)


masco:
This will miss the infrared's fixes/updates. for example, if we are using the same jump host (like ci env) the IR dir will always exist and never updated.

venkataanil:
> This will miss the infrared's fixes/updates. for example, if we are using the same jump host (like ci env) the IR dir will always exist and never updated.

Let CI call clean.yml playbook, which will remove ir dir then main.yml will install new ir dir.

bootstrap.yml
```diff
@@ -16,8 +16,19 @@
...
+        register: infrared_directory
+
+      - name: check if the infrared workspace dir exist
+        stat:
+          path: "{{ infrared_workspaces_dir }}"
+        register: infrared_workspaces_directory
+
       - name: setup infrared
         include_tasks: tasks/setup_infrared.yml
+        when: infrared_directory.stat.exists == False and infrared_workspaces_directory.stat.exists == False
```
masco: the infrared setup task called only if both the dirs are absent. but if only one dir misses the script will fail while accessing the other dir.

masco (changes requested):

bootstrap.yml
masco: TASK [run tripleo-undercloud] ****************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to change directory before execution: [Errno 2] No such file or directory: '/home/masco/infrared'"}

this is the error I am getting while infrared dir is missing.

masco: @asyedham still this condition is unchanged. i will tell you how to reproduce the above-mentioned case. on the jumphost just delete the 'infrared_dir` manually and rerun the deployment. i think we need to change the condition to `OR`.

masco (changes requested):

asyedham:
@masco Yeah, but let me explain to you how this PR works and the conditions

1. On a fresh setup, when we triggermain.yml it will run the task 
     https://github.com/redhat-performance/jetpack/blob/master/tasks/setup_infrared.yml.
2. In case because of some failure you want to re-run the main.yml playbook it skips setup_infrared.yml as both  "{{ infrared_dir 
     }}" and "{{ infrared_workspaces_dir }}" exist.
3. If you want to re-use the same ansible jump host with the latest infrared changes after a few days, you can run the cleanup.yml 
    which removes both the dirs "{{ infrared_dir }}" and "{{ infrared_workspaces_dir }}" and trigger the main.yml playbook.

So we will not encounter an issue with only one dir missing.

bootstrap.yml
masco: @venkataanil I feel we need to call this with `OR` condition. what do you think?

venkataanil: > @venkataanil I feel we need to call this with `OR` condition. what do you think?

Agree

asyedham: @masco @venkataanil I changed it OR condition can you please have look

masco:
@venkataanil @asyedham I have a different proposal for this PR. let's have a variable on group_vars. so the user can decide whether he wants an infrared dir that needs to be reused or create freshly. what you guys think?

asyedham:
> @venkataanil @asyedham I have a different proposal for this PR. let's have a variable on group_vars. so the user can decide whether he wants an infrared dir that needs to be reused or create freshly. what you guys think?

+1, we can do that but I'm curious as to why the CI job is failing

PR #301, opened 2020-11-16
OCP: use nvme for etcd

venkataanil:
This patch creates ignition configs for master node with mounting
nvme device for etcd storage. OCP installer uses these ignition
configs during cluster creation.

Issue #282, opened 2020-10-29
[Shift-on-Stack] Use the NVMe passed through for the master nodes


smalleni:
Although we are passing through NVMe devices for the OCP master nodes, it doesn't look like we are actually leveraging them.
https://github.com/openshift/installer/tree/master/docs/user/openstack#disk-requirements

Looks like we need to use controlPlane.platform.openstack.rootVolume.type and controlPlane.platform.openstack.rootVolume.size params in our install-config.yaml

smalleni:
@masco @venkataanil @akrzos FYI

PR #314, opened 2020-12-03
Use nvme for etcd for shiftstack

akrzos:


akrzos:
Addresses #282

masco:

venkataanil (commented):
Hi Alex, I have pushed this PR https://github.com/redhat-performance/jetpack/pull/301/ long back for the same. But it is not merged as not getting enough reviews. Kindly review it.

@akrzos @smalleni we are OK to merge any of PRs 301 and 314, but better to avoid duplicate efforts.

akrzos:
> Hi Alex, I have pushed this PR #301 long back for the same. But it is not merged as not getting enough reviews. Kindly review it.
> 
> @akrzos @smalleni we are OK to merge any of PRs 301 and 314, but better to avoid duplicate efforts.

Whichever one you prefer but I think the one liner in #301 will be a bit difficult to maintain or untangle when things go sideways. Just my .02 cents.

ocp_on_osp.yml
```diff
@@ -183,6 +183,43 @@
...
+      with_items:
+        - master
+        - worker
+        - bootstrap
+
+    - name: Read original master.ign config file
+      slurp:
+        src: "/home/stack/{{ ocp_cluster_name }}/master.ign.orig"
+      register: master_ign
+      when: ocp_master_etcd_nvme
```
venkataanil, line 206: where is this ocp_master_etcd_nvme defined? Can you please add it to vars file and add a description for it as a comment?

ocp_on_osp.yml
```diff
@@ -183,6 +183,43 @@
...
+
+    - name: Get master etcd on nvme ignition configuration
+      set_fact:
+        master_ign_config_base: "{{ master_ign['content'] | b64decode }}"
+        master_etcd_nvme_config: "{{ lookup('template', 'master-etcd.ign.json') }}"
```
venkataanil, line 211: Can you please submit master-etcd.ign.json file?

ocp_on_osp.yml
akrzos: The vars are [here](https://github.com/redhat-performance/jetpack/blob/13c1ac023fed5cf871c453789ad515ddf9814773/vars/shift_stack_vars.yaml#L29-L31).  I actually had them in the multiple create/delete shiftstack commit and I think rebasing merged them "out" of this commit if that makes sense.

ocp_on_osp.yml
akrzos: The file is [here](https://github.com/redhat-performance/jetpack/blob/13c1ac023fed5cf871c453789ad515ddf9814773/templates/master-etcd.ign.json).

venkataanil (commented):
Thanks Alex.

PR #318, opened 2020-12-10
[WIP] fix the issues caused when clean_nodes enabled

asyedham:
when clean_nodes enabled
- In tasks/install_os.yml, after the overcloud delete the overcloud nodes are still in cleaning state and then move to available state. So I have added a task to ensure the nodes move to available state
- In tasks/delete_introspection_failed_nodes.yml, after the introspection for the nodes to move to available state it takes longer then default deployment, so I have added a wait time of 360 seconds


asyedham:
Do not merge, I still need to test this PR

PR #333, opened 2021-01-11
create VLAN based network

### LINKED ISSUE #327: OSP install with VLAN tenant networks
We need OSP install with vlan provider networks to workaround the double encap problem when not using kuryr. We also need to test OCP install on top of this cluster.

masco:
to create a VLAN based tenant network set vlan_provider_network is true and
set the below config in extra heat config
- NeutronBridgeMappings=openshift:br-openshift,datacentre:br-ex
- NeutronNetworkVLANRanges=openshift:500:1000,datacentre:1:500

fixes #327

### DIFF
--- a/templates/compute.yaml.j2
+++ b/templates/compute.yaml.j2
@@ -274,7 +274,7 @@ resources:
                               -
                                 ip_netmask:
                                     get_param: StorageIpSubnet
-{% if tenant_interface is defined %}
+{% if tenant_interface is defined  and vlan_provider_network == false %}
                         -
                           type: ovs_bridge
                           name: br-tenant
@@ -293,6 +293,17 @@ resources:
                               -
                                 ip_netmask:
                                     get_param: TenantIpSubnet
+{% if vlan_provider_network == true and openshift_vlan_interface is defined %}
+                        -
+                          type: ovs_bridge
+                          name: br-openshift
+                          use_dhcp: false
+                          members:
+                            -
+                              type: interface
+                              name: {{ openshift_vlan_interface }}
+                              primary: true
+{% endif %}
 
 outputs:
   OS::stack_id:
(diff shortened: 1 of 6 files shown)


smalleni (commented):
@venkataanil @masco Do we not need a new bridge mapping  for he openshift_vlan_interface?

venkataanil:
> @venkataanil @masco Do we not need a new bridge mapping for he openshift_vlan_interface?

Yes

venkataanil:
@masco your PR is incomplete. It is not defining needed THT params i.e

NeutronNetworkType: vlan
network_vlan_ranges=openshift:500:1000,datacentre:1:500

templates/compute.yaml.j2
```diff
@@ -293,6 +293,17 @@ resources:
                               -
                                 ip_netmask:
                                     get_param: TenantIpSubnet
```
venkataanil, line 295: {% endif %} is missing?

templates/compute.yaml.j2
```diff
@@ -274,7 +274,7 @@ resources:
                               -
                                 ip_netmask:
                                     get_param: StorageIpSubnet
```
venkataanil, line 276: Add TenantIpSubnet here in case of vlan network

templates/compute.yaml.j2
masco: it is present at line number 287

templates/compute.yaml.j2
masco: if we set vlan_provider_network = true, it will be added here, since the below condition fails

smalleni:
@masco @venkataanil Is this working now and is the PR final or, are more changes expected?

masco:
> @masco @venkataanil Is this working now and is the PR final or, are more changes expected?

@smalleni we are facing some issues while creating an instance. working on it and will update the PR after fixing the issue.

PR #337, opened 2021-01-13
Use more than one dns server if available

akrzos:


masco:

PR #339, opened 2021-01-15
fix for handling badfish errors

asyedham:


tasks/set_boot_order_director.yml
```diff
@@ -0,0 +1,77 @@
+---
+- block:
```
masco: NIT: instead of writing the whole playbook inside the block, we can add the when condition in the caller task

masco (commented):
LGTM, except the NIT comment.

asyedham:
@masco  Can we get this merged if you are ready please?

PR #342, opened 2021-01-20
fix detecting product name when running from non-root user

asyedham:
[skip ci]
This PR fixes detecting product name for 1029u's when
running as non-root user from ansible jump host.


masco (approved):
LGTM

PR #353, opened 2021-02-04
Allow OpenShift dev preview shiftstack clusters

akrzos:


masco:

akrzos:
FYI it looks like some of the vars have changed in the install-config, working this out now - https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#minimal

akrzos:
Do you folks mind maybe checking this tomorrow so we can see if there are any changes needed?  Thanks!

vars/shift_stack_vars.yaml
```diff
@@ -6,7 +6,8 @@ osp_user_name: shiftstack_user
 # osp_project_name: "shiftstack{{ocp_cluster_index}}"
 # osp_user_name: "shiftstack{{ocp_cluster_index}}"
 
-ocp_release: 4.6.1
+ocp_release: 4.7.0-fc.5
+ocp_dev_preview: true
```
venkataanil: Will this be used by shift-on-stack CI? If so it will be good if we keep
ocp_dev_preview: false
and
ocp_release to 4.6.1 which might be stable version.

templates/install-config.yaml.j2
```diff
@@ -33,10 +33,11 @@ networking:
   type: "{{ ocp_network_type }}"
 platform:
   openstack:
+    defaultMachinePlatform:
+      type: {{ ocp_worker_flavor }}
```
venkataanil, line 37: If ocp_worker_flavor is replacement for computeFlavor, then we need a default value for ocp_worker_flavor in group_vars.

venkataanil (commented):
PLease address these comments. Thanks Alex.

vars/shift_stack_vars.yaml
akrzos: Got it, set it to those defaults.

templates/install-config.yaml.j2
akrzos: Set the default to `m4.xlarge` which is what it appears it was set to before.

PR #354, opened 2021-02-05
Add support for fc640

### LINKED ISSUE #350: support scale lab fc640
These are the new Dell devices in scale lab.
Some machines have 3 interfaces and some machines have 5 interfaces.
Need to test badfish commands and boot order in these machines.

asyedham:


venkataanil (commented):
As discussed can you please add the limitations of fc640 in the original issue which I raised and update the github  readme. You many not need these changes then.

PR #372, opened 2021-02-16
Update readme about unsupported machine types

asyedham:
[skip-ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ Currently supported distros for ansible jump host are:
 
 * > RHEL 7.7
 
+# Unsupported machine types
+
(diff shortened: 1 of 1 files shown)


asyedham:
I have updated the README with the limitation for fc640

README.md
```diff
@@ -25,6 +25,11 @@ Currently supported distros for ansible jump host are:
 
 * > RHEL 7.7
 
+# Unsupported machine types
```
masco, line 28: if the user adds the interface name in the all.yaml, it will work, right? if so you can mention that instead of saying just unsupported. so it will be helpful.

asyedham: done

PR #378, opened 2021-03-12
Jetpack support for OSP13(RHEL7.9) & OSP16.2(RHEL8.4)

pradiptapks:
Set OSP16.2 release with RHEL8.4 base OS. 

- modified:   setup_undercloud.yml
- modified:   tasks/install_os.yml

**Note**: OSP16.2 Undercloud deployed with RHEL8.4 base OS. But the overcloud still have 8.3 as the overcloud image has not yet upgraded by the OSP and it doesn't create any conflict. Also in future, it won't the Jetpack code as there are no dependencies.

### DIFF
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -40,7 +40,7 @@
     - include_tasks: tasks/update_repo.yml
       when: osp_release == 13 or 16.2
 
-      ## The 
+      ## This block of task execute when condition meet for 13 and 16.2
     - block:
         - name: update os to RHEL 7.9
           shell: |
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -9,6 +9,7 @@
         15: 8.0
         16: 8.1
         16.1: 8.2
+        16.2: 8.4
   tasks:
     - name: set async_install to empty list
       set_fact:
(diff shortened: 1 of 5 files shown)


masco:

pradiptapks:
FIX2:

- Create a separate playbook `tasks/update_repo.yml` to remove the existing repos configured by Formen during OS provision and install rhos-release-latest rpm. As this task is common in each OSP configuration, included this task in the `setup_undercloud.yml` playbook.

- Created separate tasks in `setup_undercloud.yml` to upgrade RHEL7.9 and RHEL8.4 using `rhos-release` command instead repo file. It would avoid the complexity and dependency issue.

- **Note**: OSP16.2 Undercloud deployed with RHEL8.4 base OS. But the overcloud still have 8.3 as the overcloud image has not yet upgraded by the OSP and it doesn't create any conflict. Also in future, it won't the Jetpack code as there are no dependencies.

setup_undercloud.yml
```diff
@@ -30,24 +31,36 @@
       delay: 10
       with_items: "{{ async_install }}"
 
+    - include_tasks: tasks/get_interpreter.yml
+      vars:
+        hostname: "{{ undercloud_hostname }}"
+        user: "root"
+      ignore_errors: yes
+    - include_tasks: tasks/update_repo.yml
```
asyedham, line 40: please pin this task to be executed only for  osp13 and osp16.2

asyedham (changes requested):
Please squash the commits

setup_undercloud.yml
pradiptapks: Updated

masco:
rerun tests

tasks/install_os.yml
```diff
@@ -25,10 +25,15 @@
...
+- name: set os_install before updating os RHEL 8.4
+  set_fact:
+    os_install: "RHEL 8.2"
+  when: needed_os == "RHEL 8.4"
+
 - name: set os_install for forceful provisioning
   set_fact:
     os_install: "{{ needed_os }}"
-  when: force_reprovision == true and needed_os != "RHEL 7.9"
+  when: force_reprovision == true and needed_os != "RHEL 7.9" and needed_os != "RHEL 8.4"
```
masco: NIT: the conditions can be simplified like this
``` when: force_reprovision == true and needed_os not in ["RHEL 7.9", "RHEL 8.4"]```
so it can be expandable easily in future.

File: setup_undercloud.yml [hunk 3]
```diff
@@ -30,24 +31,38 @@
...
+    - include_tasks: tasks/get_interpreter.yml
+      vars:
+        hostname: "{{ undercloud_hostname }}"
+        user: "root"
+      ignore_errors: yes
+
+    - include_tasks: tasks/update_repo.yml
+      when: osp_release == 13 or 16.2
+
     - block:
```
masco, line 44: why the single task added inside of the block?

setup_undercloud.yml
```diff
@@ -30,24 +31,38 @@
...
+             rhos-release 13 -p passed_phase2 -r 7.9;
+             yum update -y
           delegate_to: "{{ undercloud_hostname }}"
           vars:
             ansible_python_interpreter: "{{ python_interpreter }}"
             ansible_user: "root"
+      when: osp_release == 13
 
-        - name: update os to RHEL 7.9
+    - block:
```
masco: same here

masco (commented):
Thanks for adding support for OSP 16.2.
I just added some NIT comments other than that it looks good to me

tasks/install_os.yml
```diff
@@ -25,10 +25,15 @@
     os_install: "RHEL 7.8"
   when: needed_os == "RHEL 7.9"
 
+- name: set os_install before updating os RHEL 8.4
+  set_fact:
+    os_install: "RHEL 8.2"
+  when: needed_os == "RHEL 8.4"
```
asyedham, line 31: Add a condition to check if the required os is already installed, if yes then it should not define os_install fact otherwise set the os_install to required RHEL version. This will avoid re-deployment every time you trigger the main.yml

asyedham (changes requested):

PR #381, opened 2021-03-23
fix scale simulation

venkataanil:
1. Remove installing OS on hypervisors and copying keys. Let user
   do this manually. This will reduce complexity in scale
   simulation code.
2. As all infrared changes related to scale simulation are merged,
   we avoid using infrared patches in the code.

### DIFF
--- a/scale_compute_vms.yml
+++ b/scale_compute_vms.yml
@@ -35,36 +35,6 @@
          hostname_list: "{{ hostname_list|default([]) + [item.stdout] }}"
       with_items: "{{ host_list.results }}"
 
-    - name: install os
-      vars:
-        chassis_password:  "{{ instackenv_content.nodes[0].pm_password }}"
-        needed_os: "{{ (lab_name == 'scale') | ternary('CentOS 7', 'CentOS 7.7') }}"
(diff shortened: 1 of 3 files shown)


scale_compute_vms.yml
```diff
@@ -82,49 +52,11 @@
...
-
-    - name: add ssh key
-      shell: |
-        echo '{{ ansible_ssh_pass }}' | sshpass ssh-copy-id -i {{ ansible_ssh_key }} -o 'StrictHostKeyChecking no' -f root@{{ hyp }}
-      changed_when: false
-      with_items: "{{ groups.hypervisor }}"
-      loop_control:
-        loop_var: hyp
-      delegate_to: "{{ groups.hypervisor|first }}"
     - name: set rhel 7 interfaces for hypervisors
```
asyedham, line 55: can you please update the name

asyedham (changes requested):
@venkataanil I have a question will these changes also work for osp13?

scale_compute_vms.yml
venkataanil: These changes are working one. I am afraid to touch anything. I will push another change fixing these typos. Thanks :)

PR #388, opened 2021-06-23
larger external net allocation pool

venkataanil:
As most of us deploy OSP for scale testing and the default
172.17.5.0/24 is not sufficient for floating ips.
So we need to use 172.18.0.0/16 network for floating ip.

When we specify 172.18.0.0/16 in group_vars, post.yaml is
still using 172.18.0.2-172.18.0.24 as allocation pools for
floating ip network. So this patch removes the logic in
post.yml which calculates this allocation pool and instead
accept the values from group vars. User need to pass them
through group vars.


asyedham (approved):
LGTM

PR #391, opened 2021-07-01
Containerize hammercli

asyedham:
[skip-ci]
No need to maintain separate hammer host for Scale and Alias Lab
Runs hammer commands from container
Supports both the labs Scale and Alias


masco:
@asyedham is it possible to set the forman_url automatically incase if we are using scale and alias?
The PR works fine having this feature will reduce one parameter setting from the user side.

PR #394, opened 2021-07-16
Optimised Jetpack Dockerfile to reduce number of layers

sanjaychari:
There are 3 RUN commands in the Jetpack Dockerfile that each lead to the creation of one layer in the resulting image. This PR aims to reduce the number of layers created in the Jetpack image by combining the 3 RUN commands into 1 RUN command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,6 @@
 FROM centos:7
 
-RUN yum -y install git
-RUN yum -y install epel-release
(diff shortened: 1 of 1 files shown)


masco:

sanjaychari:
The order in which execution is triggered matters. Ansible should not be installed before epel-release, because it will not be found. I am closing this PR because of that.

PR #402, opened 2021-09-06
Add shift-on-stack installation instructions in README.md

sanjaychari:
This PR adds instructions to install Openshift on Openstack using Jetpack in README.md. It also adds instructions to cleanup an existing shift-on-stack deployment, along with possible issues that can be faced during deployment, and solutions to fix the issues.


masco:

Issue #426, opened 2022-01-27
Add support for compute node scale in


kemmy22:
At present we do have the facility to scale out nodes using jetpack, however the support for node scale in (compute node removal) needs to be introduced.
This will be done using a playbook called scalein.yml.

PR #428, opened 2022-02-03
Enable scale in of compute nodes

kemmy22:
Using scalein.yml playbook compute nodes can now be scaled in.
Specify the node uuids which needs to be scaled in, through a file.

Signed-off-by: Ketan Mehta <kmehta@redhat.com>


kemmy22:
closes #426

PR #437, opened 2022-04-08
Add support for OSP17 deployment

asyedham:


composable.yml
```diff
@@ -99,31 +99,7 @@
...
-      with_together:
-        - "{{ node_uuid.stdout_lines }}"
-        - "{{ node_fqdn.results }}"
-
-    - name: setting profile for node type
-      shell: |
-        source ~/stackrc
-        openstack baremetal node set {{ item.key }} --property capabilities=profile:baremetal{{ item.value }},cpu_vt:true,cpu_hugepages:true,boot_option:local,cpu_txt:true,cpu_aes:true,cpu_hugepages_1g:true,boot_mode:bios
-      loop: "{{ uuid_type_map|dict2items }}"
+        boot_mode: bios
```
sanjaychari, line 104: The ironic team mentioned that OSP 17 is leaning towards using UEFI boot mode as default. I think we might want to work with the lab team to check why UEFI doesn't work usually for supermicro hosts, so that we can use UEFI instead of BIOS in the future.

external.yml
```diff
@@ -72,3 +72,17 @@
         become: true
         ignore_errors: true
         when: external_network_vlan_id is not defined
+
+      - name: get default route
```
sanjaychari, line 81: Will we not run post.yml for OSP 17? Why has this code https://github.com/redhat-performance/jetpack/blob/master/post.yml#L6-L18 been added in this external.yml playbook again?

group_vars/all.yml
```diff
@@ -274,3 +276,6 @@ osd_objectstore: bluestore
 #introspection_timeout:
 foreman_url: https://foreman.example.com
 manila_enabled: false
+
+#image path for overcloud nodes
+remote_image_path: https://sf.redhat.com
```
sanjaychari, line 282: Do we still need this remote_image_path? I guess this was mainly because of some issues with using VLANs in overcloud networking and because of some issues with boot while using bios boot mode in older images. I think those issues might be fixed in the latest puddle now, so is this still required?

group_vars/all.yml
```diff
@@ -54,6 +55,7 @@ scale:
       5039ms:
         rhel7_interfaces:  [enp1s0f0, enp1s0f1, enp2s0f1]
         rhel8_interfaces:  [enp1s0f0, enp1s0f1, enp2s0f1]
+        rhel9_interfaces:  [enp1s0f0, enp1s0f1, enp2s0f1]
```
sanjaychari, line 59: Please add rhel9_interfaces for other hardware types too. I have templates for 5 hardware types(1029p, 1029u tn10rt, 1029u trtp, r640, 6049p) in the following link http://perfscale.perf.lab.eng.bos.redhat.com/pub/schari/osp17_250_node_scale_templates_latest/, which might be useful for reference.

introspect.yml
```diff
@@ -51,8 +51,8 @@
         vars:
           rack_info: "{{ instackenv_content.nodes[item|int].pm_addr.split('-')[1] }}"
         with_sequence: 0-{{ instackenv_content.nodes|length - 1 }}
-        when: lab_name == 'alias' and rack_info == 'e26'
+        when: lab_name == 'alias' and rack_info == 'e26' and osp_release < 17
 
       - name: enable root_hints(740xd G-CL)
         include_tasks: tasks/set_root_hints.yml
-        when: lab_name == 'alias' and hostvars['localhost']['machine_type'] == '740xd' and rack == 'e26'
+        when: lab_name == 'alias' and hostvars['localhost']['machine_type'] == '740xd' and rack == 'e26' and osp_release < 17
```
sanjaychari, line 58: Are these root hints not required for 740xd G-CL in OSP 17?

templates/ceph-storage.j2.j2
```diff
@@ -0,0 +1,35 @@
...
+
+- type: ovs_bridge
+  name: br-isolated
+  use_dhcp: false
+  members:
+  - type: interface
+    name: {{ isolated_interface }}
+    primary: true
+  - type: vlan
+    vlan_id: {{ storage_network_vlan_id }}
```
sanjaychari, line 25: I think we should use ovs bridges only for tenant and external networks, as having other networks on ovs bridges is not recommended.

sanjaychari:
I think the neutron team has plans to make RAFT clusters the default for OVN in the future, so Jetpack should also support it. Passing these parameters to the infrared deploy command enables OVN RAFT.
`--config-resource OS::TripleO::Services::OVNDBs=/usr/share/openstack-tripleo-heat-templates/deployment/ovn/ovn-dbs-cluster-ansible.yaml`

external.yml
asyedham: In OSP17  overcloud deployment  fails if the masqurade rule is not added before overcloud deploy

group_vars/all.yml
asyedham: Yes, I have tested it with the latest puddle and the issue with overcloud images still doesn't seem to fixed and during the provisioning it enters into the grub rescue mode

group_vars/all.yml
asyedham: I have added the interfaces for the hardware that I have tested OSP17. We can add the other hardware once RHEL 9 is added to foreman on Scalelab and alias lab. @sanjaychari you can create a separate PR for the hardware you have tested on

introspect.yml
asyedham: Yet to test on alias lab

composable.yml
asyedham: > I think the neutron team has plans to make RAFT clusters the default for OVN in the future, so Jetpack should also support it. Passing these parameters to the infrared deploy command enables OVN RAFT. `--config-resource OS::TripleO::Services::OVNDBs=/usr/share/openstack-tripleo-heat-templates/deployment/ovn/ovn-dbs-cluster-ansible.yaml`

This is already supported in jetpack

external.yml
sanjaychari: Shouldn't line 81 and line 88 in this file be
`when: external_network_vlan_id is defined  and osp_release < 17`
then?

PR #441, opened 2022-06-16
Add python script to generate ansible inventory

sanjaychari:
This PR adds a python script to generate ansible inventory for all overcloud hosts, or a subset of hosts, like controller or compute. This script can be run after overcloud node provisioning also, unlike tripleo-ansible-inventory which requires the overcloudrc file to be present. An example usage of this script is mentioned below.
python3 generate_ansible_inventory.py -i controller -f /home/stack/templates-output/overcloud-node-deployed.yaml -o /home/stack/custom_playbooks
This will create an ansible hosts inventory file "controller_hosts_ips" in the directory /home/stack/custom_playbooks with the controller host ctlplane IPs.
This PR has been tested for OSP 17, but should work for other versions too.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -12,3 +12,10 @@ this script is used for deploying all-in-one OSP16.1 environment
 
 Example:
   ./infrared_osp16_1.sh <hostname> full
+
+### generate_ansible_inventory.py
+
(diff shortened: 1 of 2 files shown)


asyedham:
LGTM!

PR #450, opened 2022-06-29
Fix tn10rt file

rajeshP524:


hammer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:centos8
+FROM quay.io/centos/centos:stream8
```
sanjaychari, line 1: Remove this from this PR as it is already in the other PR.

PR #451, opened 2022-06-29
fix issues with tn10rt.html file

rajeshP524:
ansible 'get_url' module sometime fails to download tn10rt.html file
so using 'shell' moduel instead. Also added the tn10rt.html as a part
of the cleanup.


asyedham (approved):
LGTM!

Issue #452, opened 2022-06-30
Use badfish container instead of badfish.py script


sanjaychari:
The official recommendation on the badfish repository is to use the badfish container[1]. They also do not plan to support the badfish.py script in the long term[2].
For this reason, Jetpack should switch to using the badfish container instead of the badfish.py script.

[1] https://github.com/redhat-performance/badfish#badfish-container
[2] https://github.com/redhat-performance/JetSki/issues/186#issuecomment-982666646

PR #454, opened 2022-07-04
Add support for containerized badfish and parallel execution

rajeshP524:
Made changes for using containerized badfish and for parallel execution
when we have list of hosts for setting boot order, badfish can take list
of all hosts at once and operate on each simultaneously.


badfish/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM quay.io/quads/badfish
+COPY bad-hosts /
```
sanjaychari: Please change the name from `bad-hosts` to `dell-hosts` or something similar. bad-hosts is a misleading name :).

common.yml
```diff
@@ -65,6 +65,27 @@
...
+            state: absent
+
+      - name: prepare dell hosts for badfish
+        shell: echo "mgmt-{{ item[1].stdout }}" >> ~/jetpack/badfish/bad-hosts
+        when: vendors is defined and item[0] == "dell"
+        with_together:
+          - "{{ vendors }}"
+          - "{{ host_list.results }}"
+
+      - name: remove existing badfish image
```
sanjaychari, line 80: If the image is up to date, I don't think we need to remove it. It will simply lead to additional time being spent. Maybe you can change this step to remove the existing image only if the image is not the latest version.

sanjaychari (changes requested):

badfish/Dockerfile
rajeshP524: I took this term from badfish repository, in which they have used it for an example. using the name 'bad-hosts' with badfish commands would be meaningful.

common.yml
rajeshP524: The existing image will contain list of dell hosts from the previous deployment. so, pulling it freshly will add the new hosts.
refer to jetpack/badfish/Dockerfile from this PR

sanjaychari: Ack

badfish/Dockerfile
sanjaychari: `bad-hosts` sounds like it's referring to hosts which are faulty/bad. `dell-hosts` or something else like `badfish-hosts` would be more appropriate.

rajeshP524: okay. will update it to 'dell-hosts'.

common.yml
```diff
@@ -65,6 +65,27 @@
         with_sequence: 0-{{ (oc_instackenv_content.nodes|length - 1) }}
         register: host_list
 
+      - name: remove existing bad-hosts file
+        file:
+            path: "~/jetpack/badfish/bad-hosts"
```
asyedham: please use a appropriate naming something like dell-hosts as bad-hosts would indicate faulty nodes

tasks/set_boot_order_director.yml
```diff
@@ -1,70 +1,35 @@
...
-    source {{ badfish_venv }}/bin/activate;
-    python3 badfish.py -H mgmt-{{ item[1].stdout }}  -u quads -p {{ chassis_password }} --clear-jobs --force
-  args:
-    chdir: "{{ ansible_user_dir }}/badfish/src/badfish"
-  when: vendors is defined and item[0] == "dell"
-  with_together:
-    - "{{ vendors }}"
-    - "{{ host_list.results }}"
+  become: yes
+  shell: podman run -it --rm quay.io/quads/badfish --host-list /bad-hosts -u quads -p {{ chassis_password }} --clear-jobs --force
```
asyedham: It would be better if you create separate PR's for badfish container and parallel execution to set boot order for readability

asyedham (changes requested):

common.yml
rajeshP524: sure

sanjaychari:
Mention that this PR resolves https://github.com/redhat-performance/jetpack/issues/452.

rajeshP524:
This PR has been split into separte PR's https://github.com/redhat-performance/jetpack/pull/457 and https://github.com/redhat-performance/jetpack/pull/459

PR #459, opened 2022-07-07
Add support for badfish parallel execution

rajeshP524:
when we have list of baremetal systems to be managed by badfish,
it can take list of all baremetal nodes at once and operate on each simultaneously

### DIFF
--- a/badfish/Dockerfile
+++ b/badfish/Dockerfile
@@ -1,0 +1,2 @@
+FROM quay.io/quads/badfish
+COPY dell-hosts /
--- a/badfish/Dockerfile
+++ b/badfish/Dockerfile
@@ -1,2 +1,0 @@
-FROM quay.io/quads/badfish
-COPY dell-hosts /
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -60,3 +60,4 @@
       - name: pull badfish image
         become: yes
         shell: podman pull quay.io/quads/badfish
+        ignore_errors: yes
--- a/common.yml
+++ b/common.yml
@@ -65,6 +65,27 @@
         with_sequence: 0-{{ (oc_instackenv_content.nodes|length - 1) }}
         register: host_list
 
+      - name: remove existing dell-hosts file
+        file:
+            path: "~/jetpack/badfish/dell-hosts"
+            state: absent
+
+      - name: prepare dell hosts for badfish
+        shell: echo "mgmt-{{ item[1].stdout }}" >> ~/jetpack/badfish/dell-hosts
+        when: vendors is defined and item[0] == "dell"
+        with_together:
+          - "{{ vendors }}"
+          - "{{ host_list.results }}"
+
+      - name: remove existing badfish image
+        shell: podman rmi -f quay.io/quads/badfish
+        become: yes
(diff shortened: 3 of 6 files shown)


common.yml
```diff
@@ -65,6 +65,27 @@
         with_sequence: 0-{{ (oc_instackenv_content.nodes|length - 1) }}
         register: host_list
 
+      - name: remove existing dell-hosts file
+        file:
+            path: "~/jetpack/badfish/dell-hosts"
```
asyedham: use "{{ playbook_dir }}/badfish/dell-hosts" as this may cause failures in ci

common.yml
```diff
@@ -65,6 +65,27 @@
...
+            state: absent
+
+      - name: prepare dell hosts for badfish
+        shell: echo "mgmt-{{ item[1].stdout }}" >> ~/jetpack/badfish/dell-hosts
```
asyedham: same here

common.yml
```diff
@@ -65,6 +65,27 @@
...
+          - "{{ vendors }}"
+          - "{{ host_list.results }}"
+
+      - name: remove existing badfish image
+        shell: podman rmi -f quay.io/quads/badfish
+        become: yes
+        ignore_errors: yes
+
+      - name: build badfish image
+        shell: podman build --tag badfish:latest -f ~/jetpack/badfish/Dockerfile
```
asyedham: same here

asyedham (changes requested):

badfish/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM quay.io/quads/badfish
+COPY dell-hosts /
```
sanjaychari: The pull rate limit issue that you are facing can be avoided by not removing and pulling the image again each time. Instead of adding `COPY dell-hosts` here and having a separate Dockerfile for badfish, you can just replace lines 80-88 in common.yml with a task that runs `podman pull quay.io/quads/badfish` if the image is not the latest version.
Then, every time when you call `podman run`, you can just use `-v dell-hosts:/dell-hosts`.
Using bind mounting on the container is much more efficient than using COPY in the Dockerfile and removing the image each time.

sanjaychari (changes requested):

badfish/Dockerfile
rajeshP524: ack

common.yml
rajeshP524: ack

bootstrap.yml
```diff
@@ -60,3 +60,9 @@
       - name: pull badfish image
         become: yes
         shell: podman pull quay.io/quads/badfish
+        ignore_errors: yes
+
+      - name: set SELinux to Permissive
```
asyedham: use ansible in-built module to put SELinux in permissive mode

asyedham (changes requested):

bootstrap.yml
rajeshP524: as ansible in-built module requires a reboot to get changes reflected, I have removed this task and added it as a requirement in readme file

asyedham (approved):
LGTM!

Issue #438, opened 2022-04-21
Change OSP 16 ML2/OVN deployments to DVR by default


sanjaychari:
OSP 16/Train with ML2/OVN has DVR enabled by default.
https://github.com/openstack/tripleo-heat-templates/blob/stable/train/deployment/neutron/neutron-plugin-ml2-ovn.yaml#L183
Jetpack should also have this change.

karelyatin:
Also found a similar issue[1] which is closed with reference to https://github.com/redhat-performance/jetpack/pull/167 but doesn't look that closes it. Also there was a question if extra_templates option in group_vars/all.yml to pass the DVR template can be used.

[1] https://github.com/redhat-performance/jetpack/issues/166

PR #461, opened 2022-07-21
Change Jetpack deployments to DVR by default

rajeshP524:
This PR resolves issue https://github.com/redhat-performance/jetpack/issues/438 and https://github.com/redhat-performance/jetpack/issues/496

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -208,7 +208,9 @@ vm_image_url:
    rhel8: http://download.hosts.prod.upshift.rdu2.redhat.com/released/RHEL-8/8.1.0/BaseOS/x86_64/images/rhel-guest-image-8.1-263.x86_64.qcow2
    rhel8_2: http://download.eng.pek2.redhat.com/released/RHEL-8/8.2.0/BaseOS/x86_64/images/rhel-guest-image-8.2-290.x86_64.qcow2
 #Neutron backend, if not set default will be used.
-#neutron_backend: ovn
+#we are not supporting ML2/OVS for DVR as of now, as it needs
+#modification of default roles_data.yaml
+neutron_backend: ovn
 #ocp installation
 shift_stack: false
 vlan_provider_network: false
@@ -290,3 +292,7 @@ osds_per_device: 1
 #overcloud_timeout:
 foreman_url: https://foreman.example.com
 manila_enabled: false
+
+# DVR is enabled by default for osp_releases >= 16, set this to false for non-dvr based deployments.
+# This variable has no affect on osp_releases < 16, as jetpack doesn't support dvr for osp_releases < 16 as of now.
+dvr_enabled: true
--- a/templates/network-environment_v2.yaml.j2
+++ b/templates/network-environment_v2.yaml.j2
@@ -27,10 +27,17 @@ parameter_defaults:
     ControllerParameters:
       NeutronBridgeMappings: "datacentre:br-ex,tenant:br-isolated"
 {% if composable_roles %}
+{% if dvr_enabled == false %}
 {% for node_type in machine_types %}
     Compute{{ node_type }}Parameters:
       NeutronBridgeMappings: ""
 {% endfor %}
 {% endif %}
+{% endif %}
     NeutronExternalNetworkBridge: ""
     NeutronNetworkVLANRanges: " datacentre:300:900, tenant:200:2000"
+
+{% if dvr_enabled == true %}
+    NeutronBridgeMappings: "datacentre:br-ex,tenant:br-tenant"
+    NeutronEnableDVR: 'true'
+{% endif %}
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -280,9 +280,14 @@
           deployment_timeout: 2400
         when: scale_compute_vms == true
 
+      - name: set fact for deployment with DVR
+        set_fact:
+          dvr_params: "--network-dvr yes"
+        when: dvr_enabled and osp_release|int >= 16
+
       - name: run tripleo-overcloud deploy
         shell: |
             source .venv/bin/activate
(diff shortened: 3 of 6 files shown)


group_vars/all.yml
```diff
@@ -280,3 +280,8 @@ manila_enabled: false
 
 #image path for overcloud nodes
 remote_image_path: https://sf.redhat.com
+
+#DVR is the default for ML2/OVN deployments for osp releases >= 16
+#change below variable to false to disable DVR
+#Note: if Neutron_backend == 'ovs' or osp_release < 16, then DVR won't be enabled irrespective of setting below variable to true/false
```
sanjaychari: Do not have this feature. DVR is supported for ML2/OVS also, and is also supported in OSP 13. So change line 335 in overcloud.yml to only check for dvr_enabled == false.
Also change line 344 in overcloud.yml similarly.

sanjaychari (changes requested):

group_vars/all.yml
rajeshP524: > Do not have this feature. DVR is supported for ML2/OVS also, and is also supported in OSP 13. So change line 335 in overcloud.yml to only check for dvr_enabled == false. Also change line 344 in overcloud.yml similarly.

ack

pradiptapks:
@rajeshP524 @sanjaychari @asyedham 

Please find my below suggestion:

- There are a couple of additional changes needed in DVR mapping via Jetpack.
- The compute interface template should rename br-ex to another name "br-dvr". 
- Advantage: When the External (ingress/egress) traffic is processed, the traffic should not be broadcast to the controller br-ex bridge.
- To do this Jetpack has to replace br-ex bridge to br-dvr in compute the interface template
```
-
  type: ovs_bridge
  name: br-dvr
  use_dhcp: false
  members:
    -
      type: interface
      name: ens7f1
      primary: true
```

- The network environment should have a correct mapping to the bridge.
```
  NeutronNetworkVLANRanges:
    - dvr:410:410
  NeutronBridgeMappings:
    - dvr:br-dvr
```
- A VLAN gateway route should be created in Undercloud for DVR network reachability. Example:
```
@undercloud ~]$ ip addr show ens7f1.410
12: ens7f1.410@ens7f1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 3c:fd:fe:c5:01:11 brd ff:ff:ff:ff:ff:ff
    inet 192.168.4.1/24 scope global ens7f1.410
       valid_lft forever preferred_lft forever

```

group_vars/all.yml
pradiptapks: @sanjaychari @rajeshP524 

Before any integration, you can check Red Hat OpenStack Platform Life Cycle document for OSP release support

As per the data, OSP13 has maintenance support already expired. I don't think we need any major scale or performance test plan to execute in OSP13 release cycle. So we can focus on OSP16.x and 17.x release.


https://access.redhat.com/support/policy/updates/openstack/platform

sanjaychari:
@pradiptapks 

I had a query. As far as I understand, for east-west traffic and north-south traffic with floating IPs in IPv4; DVR will only route traffic through the compute nodes and will not broadcast traffic to br-ex on the controller node, irrespective of whether br-ex or another bridge is used on the compute node. For north-south traffic without floating IP, traffic has to travel through the controller node anyway.
In this case, what would be the advantage of having a separate br-dvr bridge on compute nodes?

rajeshP524:
> @rajeshP524 @sanjaychari @asyedham
> 
> Please find my below suggestion:
> 
> * There are a couple of additional changes needed in DVR mapping via Jetpack.
> * The compute interface template should rename br-ex to another name "br-dvr".
> * Advantage: When the External (ingress/egress) traffic is processed, the traffic should not be broadcast to the controller br-ex bridge.
> * To do this Jetpack has to replace br-ex bridge to br-dvr in compute the interface template
> 
> ```
> -
>   type: ovs_bridge
>   name: br-dvr
>   use_dhcp: false
>   members:
>     -
>       type: interface
>       name: ens7f1
>       primary: true
> ```
> 
> * The network environment should have a correct mapping to the bridge.
> 
> ```
>   NeutronNetworkVLANRanges:
>     - dvr:410:410
>   NeutronBridgeMappings:
>     - dvr:br-dvr
> ```
> 
> * A VLAN gateway route should be created in Undercloud for DVR network reachability. Example:
> 
> ```
> @undercloud ~]$ ip addr show ens7f1.410
> 12: ens7f1.410@ens7f1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
>     link/ether 3c:fd:fe:c5:01:11 brd ff:ff:ff:ff:ff:ff
>     inet 192.168.4.1/24 scope global ens7f1.410
>        valid_lft forever preferred_lft forever
> ```

Thanks @pradiptapks for these suggestions.
Leaving the name as "br-ex" on compute nic template as "br-ex" is removed from controller.

group_vars/all.yml
```diff
@@ -207,7 +207,7 @@ vm_image_url:
    rhel8: http://download.hosts.prod.upshift.rdu2.redhat.com/released/RHEL-8/8.1.0/BaseOS/x86_64/images/rhel-guest-image-8.1-263.x86_64.qcow2
    rhel8_2: http://download.eng.pek2.redhat.com/released/RHEL-8/8.2.0/BaseOS/x86_64/images/rhel-guest-image-8.2-290.x86_64.qcow2
 #Neutron backend, if not set default will be used.
-#neutron_backend: ovn
+neutron_backend: ovn
```
sanjaychari, line 213: Add a comment that we are not supporting ml2/ovs for DVR right now because it needs changes in roles_data.yaml file.

sanjaychari (changes requested):
Remove the swp file also.

group_vars/all.yml
rajeshP524: done, added!

rajeshP524:
> Remove the swp file also.

done

sanjaychari (approved):

pradiptapks (approved):
Regarding the control-plane network in `br-ex`, I had a syncup with Rajesh and Sanjay, and the mapping is based on the condition basis on which the node has 3 interfaces.

templates/network-environment_v2.yaml.j2
```diff
@@ -27,10 +27,17 @@ parameter_defaults:
...
       NeutronBridgeMappings: "datacentre:br-ex,tenant:br-isolated"
 {% if composable_roles %}
+{% if dvr_enabled == false %}
 {% for node_type in machine_types %}
     Compute{{ node_type }}Parameters:
       NeutronBridgeMappings: ""
 {% endfor %}
+{% endif %}
 {% endif %}
     NeutronExternalNetworkBridge: ""
```
asyedham, line 37: for dvr NeutronExternalNetworkBridge: "br-ex"?

PR #463, opened 2022-08-04
Add support for composable roles for OSP17

asyedham:


tosky:
Out of curiosity, have you tried using the new generate_baremetal_deployment_from_roles_data.py script which is used by default by infrared in the composable role case, and which may help reducing some duplication (by relying on some files provided by THT)?

asyedham:
@tosky That's interesting, we haven't used this  new generate_baremetal_deployment_from_roles_data.py script, is there any example/ doc how this creates the composable roles as in our  case the role is create based on hardware type like Compute1029u,Computer630, Compute1029p so on

composable_prepare_nic_configs.yml
```diff
@@ -243,3 +263,60 @@
...
+            delegate_to: undercloud
+            with_together:
+              - "{{ machine_types }}"
+              - "{{ ctlplane_interfaces }}"
+              - "{{ external_interfaces }}"
+              - "{{ tenant_interfaces }}"
+              - "{{ isolated_interfaces }}"
+              - "{{ openshift_vlan_interfaces }}"
+
+        when: osp_release|int == 17
```
masco: can you use >= 17, it won't work on the later versions. it will be good to change all the == 17 to >= 17

masco (changes requested):

composable_prepare_nic_configs.yml
asyedham: Done

masco: still i am seeing ==

PR #464, opened 2022-08-18
Restore SELinux mode

rajeshP524:
Jetpack sets selinx mode to Permissive, This PR restores that change

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -62,6 +62,10 @@
         shell: podman pull quay.io/quads/badfish
         ignore_errors: yes
 
+      - name: get SELinux status
+        shell: getenforce
+        register: selinux_status
+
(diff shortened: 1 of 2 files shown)


bootstrap.yml
```diff
@@ -62,6 +62,16 @@
         shell: podman pull quay.io/quads/badfish
         ignore_errors: yes
 
+      - name: get SELinux status
+        shell: getenforce
+        register: selinux_status
+
+      - set_fact: se_enforcing=false
```
sanjaychari: I don't think a separate fact is required here. Just check directly for `when: '"Enforcing" in selinux_status.stdout'`
in line 75 in `post.yml`.

sanjaychari (changes requested):

PR #466, opened 2022-09-09
Increase OCP ClusterNetwork cidr range

sanjaychari:
For installing OCP clusters with large number of worker nodes, the CIDR range for the cluster network has to be changed from /14 to /13.


masco:
LGTM

PR #467, opened 2022-09-12
Flavor ram size

jtaleric:
fixing what appears to be a typo

Signed-off-by: Joe Talerico (rook) <joe.talerico@gmail.com>


masco (approved):
LGTM

PR #471, opened 2022-09-16
setting the boot mode to director (dell)

asyedham:
This PR creates a separate playbook for setting boot mode to director


rajeshP524:
This PR saves us some time when using jetpack with tags, LGTM!

PR #473, opened 2022-09-16
Composable roles: Fix the node count when nodes fails introspection

asyedham:
This PR fixes
failed_nodes_machine_type fact
Adjusts node count for computes after introspection fails
Optimize the code to run faster with overcloud tag

### DIFF
--- a/post_introspect.yml
+++ b/post_introspect.yml
@@ -1,6 +1,12 @@
 ---
 - hosts: undercloud
+  vars:
+      tn10rt: "{{ hostvars['localhost']['tn10rt'] }}"
   tasks:
+    - name: check and delete introspection failed nodes
+      include_tasks: tasks/delete_introspection_failed_nodes.yml
+      when: scale_compute_vms == false
+
     - name: get all nodes
       shell: |
         source ~/stackrc
@@ -32,6 +38,7 @@
               fi
             done
           register: storage_disks_info_comp
+
           with_items: "{{ total_nodes.stdout_lines | default([]) }}"
           when: composable_roles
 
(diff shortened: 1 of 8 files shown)


post_introspect.yml
```diff
@@ -53,8 +60,83 @@
...
-    - name: remove the templates dir
-      file:
-        state: absent
-        path: /home/stack/templates
-      ignore_errors: true
+    - name: setting node capabilities (boot_mode bios)
+      vars:
+        boot_mode: bios
+      include_tasks: tasks/set_boot_mode.yml
+      when: osp_release|int == 17 or composable_roles
```
sanjaychari: I think `osp_release|int >= 17` can be checked instead?

tasks/set_root_hints.yml
```diff
@@ -7,7 +7,7 @@
   with_items:
     - "{{ total_nodes.stdout_lines }}"
 
-- name: set root hints based on model (740XD G-CL)
+- name: set root hints based on model
```
sanjaychari, line 10: Are we setting root hints for all models here? If that's the case, maybe `name: set root hints` is a better name?

post_introspect.yml
asyedham: Done

post_introspect.yml
```diff
@@ -53,8 +60,83 @@
...
+      vars:
+        boot_mode: bios
+      include_tasks: tasks/set_boot_mode.yml
+      when: osp_release|int >= 17 or composable_roles
+
+    - name: get overcloud node images
+      vars:
+        image_path: /home/stack/images
+      include_tasks: tasks/overcloud_images.yaml
+      when: osp_release|int == 17
```
sanjaychari, line 73: Can you change this also to  `when: osp_release|int >= 17`?

asyedham: As we are facing the issue with overcloud images for only osp 17, the condition i have added make sense. Thoughts?

sanjaychari: Yeah, makes sense.

sanjaychari:
LGTM

PR #477, opened 2022-09-20
Added boot_mode params for setting boot_mode to bios at the time of undercloud install.

kemmy22:
Added boot_mode params for setting boot_mode to bios at the time of undercloud install.

### DIFF
--- a/undercloud.yml
+++ b/undercloud.yml
@@ -54,3 +54,17 @@
         args:
             chdir: "{{ infrared_dir }}"
         changed_when: false
+        when: osp_release|int < 17
+      - name: run tripleo-undercloud for osp17
+        shell: |
+          source {{ infrared_dir }}/.venv/bin/activate
+          infrared tripleo-undercloud -vv \
+          --version {{ osp_release }} \
+          --build {{ osp_puddle }} \
+          --images-task rpm \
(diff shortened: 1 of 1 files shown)


pradiptapks:
@kemmy22 Is there any specific reason we are depending `ilo` driver instead of `ipmi`.

undercloud.yml
```diff
@@ -50,6 +50,7 @@
           --version {{ osp_release }} \
           --build {{ osp_puddle }} \
           --images-task rpm \
+          --hieradata-config "ironic::conductor::default_boot_mode=bios,ironic::drivers::ilo::default_boot_mode=bios" \
```
pradiptapks, line 65: `ilo` driver is specific to HPE based server. Is there any reason we don't use `ipmi`?

sanjaychari, line 65: It might be a good idea to add a condition check for `osp_release>=17` here, just in case the line interferes with older releases.

kemmy22: It refers to the section in ironic.conf, default_boot_mode is mentioned at 2 places in the file.

kemmy22: although this is infra specific, so should be the same for 13, 16 as well but it makes sense to add the release check. i'll add that.

sanjaychari:
LGTM

PR #481, opened 2022-09-23
Fix Containerized Jetpack

rajeshP524:
jetpack uses hammer and badfish in a container, So this patch adds support for "podman inside a podman". Also updated the image to centos:stream8 and fixed some issues with old Dockerfile

### DIFF
--- a/jetpack_container.sh
+++ b/jetpack_container.sh
@@ -3,9 +3,9 @@
 ansible_dir=/jetpack
 
 # build jetpack container
-podman build -t jetpack .
+sudo podman build -t jetpack .
 
 # run jetpack container
-podman run -it \
+sudo podman run --privileged -it \
       -v ./group_vars/all.yml:$ansible_dir/group_vars/all.yml:Z \
       -t localhost/jetpack
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,50 @@
-FROM centos:7
+FROM quay.io/centos/centos:stream8
 
-RUN yum -y install git
-RUN yum -y install epel-release
-RUN yum -y install ansible
+# Install requirements.
+RUN yum -y update \
+ && yum -y install \
+      epel-release \
+      sudo \
+      python3 \
+      python3-pip \
+      sshpass \
+ && yum clean all
+
+# Don't include container-selinux and remove
+# directories used by yum that are just taking
+# up space.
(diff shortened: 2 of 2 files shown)


jetpack_container.sh
```diff
@@ -3,9 +3,9 @@
 ansible_dir=/jetpack
 
 # build jetpack container
-podman build -t jetpack .
+sudo podman build -t jetpack .
 
 # run jetpack container
-podman run -it \
+sudo podman run --privileged -it \
```
sanjaychari, line 9: Why does podman need to run in privileged mode?

sanjaychari: Maybe try something like this instead?

`podman run --cap-add=sys_admin,mknod --device=/dev/fuse --security-opt label=disable quay.io/podman/stable podman run ubi8-minimal echo hello`

Source : https://www.redhat.com/sysadmin/podman-inside-container

rajeshP524: > Why does podman need to run in privileged mode?

--privileged is needed to run podman inside a container

rajeshP524: > Maybe try something like this instead?
> 
> `podman run --cap-add=sys_admin,mknod --device=/dev/fuse --security-opt label=disable quay.io/podman/stable podman run ubi8-minimal echo hello`
> 
> Source : https://www.redhat.com/sysadmin/podman-inside-container

this might work without --privileged. will give it a try

rajeshP524: > Maybe try something like this instead?
> 
> `podman run --cap-add=sys_admin,mknod --device=/dev/fuse --security-opt label=disable quay.io/podman/stable podman run ubi8-minimal echo hello`
> 
> Source : https://www.redhat.com/sysadmin/podman-inside-container

I did hit some errors with this change

sanjaychari: Then I think it might be okay to use --privileged.

sanjaychari: Add a comment in README.md saying that jetpack container will use `--privileged` flag.

rajeshP524: > Add a comment in README.md saying that jetpack container will use `--privileged` flag.

ack

asyedham, line 9: why do you need sudo?

asyedham (changes requested):

PR #484, opened 2022-10-12
fixes for scale deployment

asyedham:
- support setting root_hints for storage class servers
- create custom roles for compute with available machine types
- minor fixes to support ci


composable.yml
```diff
@@ -7,6 +7,17 @@
         infrared_oc_deploy_template:  "{{ ansible_user_dir }}/.infrared/plugins/tripleo-overcloud/templates/overcloud_deploy.sh.j2"
       shell: |
         cp   {{ oc_deploy_template }} {{ infrared_oc_deploy_template }}
+
+    - name: avoid compute role of controller machine type
```
sanjaychari: If I understand correctly, this task will only run when the number of nodes of this hardware type is equal to the expected number of controllers? If yes, then maybe the task name can be changed accordingly to avoid confusion?

sanjaychari:
LGTM

overcloud.yml
```diff
@@ -51,6 +51,13 @@
         path: /home/stack/templates
       ignore_errors: true
 
+    - name: workaround for BZ 2123252
+      become: true
+      shell: |
+        dnf install containernetworking-plugins -y
+        sed -i 's/netavark/cni/g' /etc/containers/containers.conf
+      when: osp_release|int == 17
```
masco, line 59: use >= 17

post_introspect.yml
```diff
@@ -70,7 +70,7 @@
       vars:
         image_path: /home/stack/images
       include_tasks: tasks/overcloud_images.yaml
-      when: osp_release|int == 17
+      when: remote_image_path is defined and osp_release|int == 17
```
masco: same as above

overcloud.yml
asyedham: we are hitting this bug on osp 17, so i have added workaround for now till the fix is released

post_introspect.yml
asyedham: this was deprecated in the PR https://github.com/redhat-performance/jetpack/pull/487 as now we are using overcloud-full image in bios mode

masco (approved):
LGTM

PR #485, opened 2022-10-13
Restructure ocp_on_osp

rajeshP524:
With this PR, ocp_on_osp.yml can be used for installing single and multiple openshift clusters. Restructured code in such a way that, it will be easier to integrate it with browbeat as a workload.


ocp_on_osp.yml
```diff
@@ -1,309 +1,42 @@
...
-        dest: /home/stack/.kube/config
-        mode: a+rx
-        remote_src: yes
-
-- import_playbook: create_ocp_infra_nodes.yml
-  when: (ocp_create_infra_nodes|bool == true or ocp_create_workload_nodes|bool == true)
+    - name: Add undercloud host to inventory for each cluster to be created
+      add_host:
+        name: "undercloud-{{ item }}"
+        group: underclouds
```
sanjaychari: Is there no better way to do this? Having a group called underclouds with the same host repeated multiple times doesn't seem optimal.

tasks/set_masquerade.yml
```diff
@@ -0,0 +1,12 @@
...
+  shell: sudo iptables --flush
+
+- name: get default route
+  shell: |
+    sudo ip r | grep default | cut -d ' ' -f5
+  register: default_route
+
+- name: masquerade on public interface
+  shell: |
+    sudo iptables -t nat -A POSTROUTING -o {{ default_route.stdout }} -j MASQUERADE
```
sanjaychari, line 12: `sudo iptables --flush` will not flush rules in the nat table, as far as I have observed.
Is it required to add this rule again here? It will lead to the same rule being added multiple times.

rajeshP524: This rule will not be added multiple times, we can customize it in shift_stack_vars now. 
https://github.com/redhat-performance/jetpack/pull/485/commits/e501c7f4d1cdc1e4959d6bff12c4afa63a33d6bf#diff-f9144ea1aa91b6992a70c42069fbd7414958b637270934198cbf3d5d2a25681aR17-R18

rajeshP524: > `sudo iptables --flush` will not flush rules in the nat table, as far as I have observed.

Even I see that, rules in nat are not flushed with this. but the old playbook[1] has this task.. so i preserved the same
[1] https://github.com/redhat-performance/jetpack/blob/master/ocp_on_osp.yml#L11

sanjaychari: Ack

ocp_on_osp.yml
rajeshP524: > Is there no better way to do this? Having a group called underclouds with the same host repeated multiple times doesn't seem optimal.

I think this can further be optimized by overriding fork and serial values

ocp_cleanup.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Determine undercloud and allocate inventory
+  hosts: localhost
+  gather_facts: false
+  vars_files:
+    - vars/shift_stack_vars.yaml
+  tasks:
+    - name: list openstack projects for installed clusters
+      shell: |
+        ansible all -i '{{ undercloud_host }},' -m shell -a 'source /home/stack/overcloudrc && openstack project list -c Name -f value | grep rhocp' -u stack | grep rhocp
```
asyedham: instead of using the adhoc command can you run this task on undercloud using the shell module

ocp_on_osp.yml
asyedham: can't we just have group undercloud and use it for creating each cluster?

tasks/create_ocp_infra_nodes.yml
```diff
@@ -0,0 +1,121 @@
+---
+- name: import shift_stack_vars data
+  include_vars:
+    dir: "{{ playbook_dir }}/vars"
+    files_matching: shift_stack_vars.yaml
+
+- name: Get cluster name
+  shell: |
+    {%raw%}oc get machineset -n openshift-machine-api -o=go-template='{{ (index (index .items 0).metadata.labels {%endraw%} "{{ machineset_metadata_label_prefix }}/cluster-api-cluster" {%raw%}) }}'{%endraw%}
```
asyedham, line 9: why are using {%raw%}?

asyedham (changes requested):

ocp_on_osp.yml
rajeshP524: > can't we just have group undercloud and use it for creating each cluster?

We have only one group called 'underclouds' under which same undercloud_host will be repeated. And this inventory is prepared dynamically.
This is needed when installing multiple openshift clusters.

tasks/create_ocp_infra_nodes.yml
rajeshP524: > why are using {%raw%}?

This task comes from original playbook[1]
[1] https://github.com/redhat-performance/jetpack/blob/master/create_ocp_infra_nodes.yml#L13

ocp_cleanup.yml
rajeshP524: > instead of using the adhoc command can you run this task on undercloud using the shell module

This task runs on jumphost.., running it on undercloud directly using shell module needs a manual intervention to prepare inventory file.
we already have undercloud_host[1] from shift_stack_vars.yaml
used adhoc command, avoiding to create a special inventory for this.
[1] https://github.com/redhat-performance/jetpack/pull/485/commits/f3a0f45abb2d5ed0f87832b9e25a17223adf42d3#diff-f9144ea1aa91b6992a70c42069fbd7414958b637270934198cbf3d5d2a25681aR10

ocp_on_osp.yml
rajeshP524: > Is there no better way to do this? Having a group called underclouds with the same host repeated multiple times doesn't seem optimal.

@sanjaychari 
The default forks value is 5, which means ansible maintains only 5 parallel ssh connections by default.
By overriding fork value to desired count, we can achieve parallelism.. which should be optimal as any other method.
This method also has an advantage when we want to install openshift clusters in batches.

Example usage:
ansible-playbook -vvv -f 25 ocp_on_osp.yml

asyedham: any update on this?

asyedham: As discussed @rajeshP524  will come up with better naming for group `underclouds` which is misleading

rajeshP524: sure @asyedham

rajeshP524: updated the name

masco (approved):
LGTM

sanjaychari (approved):

asyedham (approved):
LGTM!

PR #486, opened 2022-10-14
Extends default partitions for overcloud nodes

### LINKED ISSUE #483: Extend default partition for overcloud nodes during provisioning
By default overcloud nodes gets provisioned with a very small disk space allocated to / , /var/ and /tmp which won't be sufficient for performing regular overcloud operations on the cloud.

kemmy22:
It uses a custom overcloud-node-growvols playbook to extend the partitions.
This PR resolves https://github.com/redhat-performance/jetpack/issues/483

### DIFF
--- a/overcloud-node-growvols.yml
+++ b/overcloud-node-growvols.yml
@@ -1,0 +1,13 @@
+---
+- name: Specify extra vars for extending logical volumes
+  import_playbook: /usr/share/ansible/tripleo-playbooks/cli-overcloud-node-growvols.yaml
+  vars:
+    growvols_args: >
+        /=30GB
+        /tmp=5GB
+        /home=10GB
+        /var/log=10GB
+        /var/log/audit=1GB
+        /var=90%
+        /srv=10%
+
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -258,6 +258,16 @@
           deployment_timeout: 2400
         when: scale_compute_vms == true
 
+      - name: change the default values for overcloud-node-growvols
+        copy:
+          src: "overcloud-node-growvols.yml"
+          dest: "/home/stack/cli-overcloud-node-growvols.yml"
+        delegate_to: "{{ undercloud_hostname }}"
+        vars:
+          ansible_python_interpreter: "{{ python_interpreter }}"
+          ansible_user: "stack"
+        when: osp_release|int >= 17
+
       - name: run tripleo-overcloud deploy
         shell: |
             source .venv/bin/activate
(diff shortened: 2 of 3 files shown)


overcloud-node-growvols.yml
```diff
@@ -0,0 +1,82 @@
+---
```
sanjaychari, line 1: This file seems to have code similar to https://github.com/openstack/tripleo-ansible/blob/master/tripleo_ansible/playbooks/cli-overcloud-node-growvols.yaml with parameter values changed. In order to make it easier to maintain in the future, I think it would be better to have a task which replaces the required values in `/usr/share/ansible/tripleo-playbooks/cli-overcloud-node-growvols.yaml` directly instead of storing this file in Jetpack and copying it later.

sanjaychari (changes requested):

overcloud-node-growvols.yml
kemmy22: Yes.
Actually, that's also not the correct way of doing it. I was going through the upstream docs and we can make use of extra_vars to achieve this. I'll amend the commit in a while, testing this with that approach.

overcloud-node-growvols.yml
```diff
@@ -0,0 +1,13 @@
...
+- name: Specify extra vars for extending logical volumes
+  import_playbook: /usr/share/ansible/tripleo-playbooks/cli-overcloud-node-growvols.yaml
+  vars:
+    growvols_args: >
```
sanjaychari, line 5: Will this set the same values for all roles?

kemmy22: yes, at the moment we're not segregating based on role type as the bump up in space is not too large. If we segregate currently based on role it might increase complexity. We'll have to specify different playbooks for role type in that case.

sanjaychari: Okay, if there is no specific need for role based segragation, this should be fine.

sanjaychari:
LGTM

overcloud.yml
```diff
@@ -258,6 +258,15 @@
           deployment_timeout: 2400
         when: scale_compute_vms == true
 
+      - name: change the default values for overcloud-node-growvols
```
sanjaychari, line 261: I think it would be good to add a condition check for `when: osp_release|int >=17` on this task as the growvols playbook is not there in older OSP releases.

kemmy22: ack, i'll make the change

sanjaychari:
LGTM

sanjaychari (approved):

masco:
@kemmy22 can you rebase this PR?

kemmy22:
@masco Resolved Merge Conflicts.

masco (approved):

PR #487, opened 2022-10-18
use overcloud-full image

asyedham:
enable usage of overcloud-full images in OSP17+


post_introspect.yml
```diff
@@ -66,19 +66,6 @@
       include_tasks: tasks/set_boot_mode.yml
       when: osp_release|int >= 17 or composable_roles
 
-    - name: get overcloud node images
-      vars:
-        image_path: /home/stack/images
-      include_tasks: tasks/overcloud_images.yaml
-      when: osp_release|int == 17
-
-    - name: Set name for overcloud nodes
```
sanjaychari, line 75: I think it would be useful to set the name for overcloud nodes. Has this task been moved to another playbook already? If not, why is this task being removed?

asyedham: since the instackenv.json has name field infrared set the name using it. We no longer need to explicitly set the name to overcloud nodes

sanjaychari: Ack

sanjaychari:
LGTM

masco (approved):
LGTM

PR #491, opened 2022-10-20
Fix the node count when introspection fails osp < 17

### LINKED ISSUE #474: Fix the node count when introspection  fails osp < 17
The node count needs to be adjusted according when the introspection fails
https://github.com/redhat-performance/jetpack/blob/master/templates/nodes_data.yml.j2#L9

asyedham:
fixes #474


sanjaychari:
LGTM

masco (approved):
LGTM

PR #493, opened 2022-10-21
Run hammer in a rootless container

rajeshP524:
1) hammer doesn't need root privileges to update the host OS. 
2) run hammer container with --rm option as its no more required for the further run.
   Also this avoids containers ending up in an exited state on jumphost/ansible_control_node


sanjaychari (approved):
LGTM

masco (approved):
LGTM

PR #497, opened 2022-10-29
Implements osd-spec for specifying storage_disks

### LINKED ISSUE #489: Rework storage_node_disks for OSP17
The process for specifying ceph specific parameters with OSP deployment like name of block devices, pg count, etc have changed with ceph5 and now relies on cephadm params.

kemmy22:
Ceph 5 with OSP17 now makes use of a osd-spec file to specify devices on which OSDs will be created.
Also includes code optimisation for reducing the time to overcloud deployment.
This change implements osd-spec and resolves https://github.com/redhat-performance/jetpack/issues/489

### DIFF
--- a/overcloud.yml
+++ b/overcloud.yml
@@ -257,13 +257,25 @@
 
       - name: set facts for ceph deployment
         set_fact:
-          ceph_params: "--storage-backend ceph --storage-external no"
+          ceph_params: "{{ (osp_release|int >= 17) | ternary('--storage-backend ceph --storage-external no --ceph-osd-spec-file {{ osd_spec_file }}','--storage-backend ceph --storage-external no') }}"
         when: ceph_enabled
 
       - name: set deployment_timeout to 2400
         set_fact:
           deployment_timeout: 2400
         when: scale_compute_vms == true
+
+      - block:
+        - name: set fact for osd-spec.yml file
+          set_fact:
+            osd_spec_file: "{{ ansible_user_dir }}/osd-spec.yml"
+
+        - name: generate osd-spec.yml.j2
+          template:
+            src: osd-spec.yml.j2
+            dest: "{{ osd_spec_file }}"
+            force: yes
+        when: ceph_enabled and osp_release|int >= 17
 
       - name: run tripleo-overcloud deploy
         shell: |
--- a/post_introspect.yml
+++ b/post_introspect.yml
@@ -27,7 +27,6 @@
           with_items: "{{ total_nodes.stdout_lines | default([]) }}"
           when: composable_roles != true
 
-
         - name: introspection data for a nodes
           shell: |
             source ~/stackrc
@@ -53,12 +52,12 @@
             storage_node_disks: "{{ storage_node_disks|default([]) + item.stdout_lines }}"
           with_items: "{{ storage_disks_info_comp.results }}"
           when: composable_roles
-      when: storage_node_disks is not defined and ceph_enabled
+      when: storage_node_disks is not defined and ceph_enabled and osp_release|int < 17
 
     - name: show the storage_node_disks
       debug:
         msg: "{{ storage_node_disks|unique }}"
-      when: ceph_enabled
+      when: ceph_enabled and osp_release|int < 17
 
     - name: setting node capabilities (boot_mode bios)
       vars:
@@ -119,7 +118,7 @@
       - name: set fact for storage nodes
         set_fact:
           storage_node_disks: "{{ hostvars['undercloud']['storage_node_disks']|unique }}"
-        when: ceph_enabled and storage_node_disks is not defined
+        when: ceph_enabled and storage_node_disks is not defined and osp_release|int < 17
 
       - name: generate internal.yml.j2
         template:
--- a/templates/osd-spec.yml.j2
+++ b/templates/osd-spec.yml.j2
@@ -1,0 +1,10 @@
+data_devices:
+{%if storage_node_disks is defined %}
+  paths:
+{% for disk in storage_node_disks %}
+    - '{{disk}}'
+{% endfor %}
+{% else %}
+  all: true
+{% endif %}
+osds_per_device: {{ osds_per_device }}
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -279,6 +279,9 @@ osd_objectstore: bluestore
 #Note:By default storage_node_disks can be detected automatically
 #using introspection data
 #storage_node_disks: ['/dev/nvme0n1']
+#Note: By default the number of osds_per_device is set to 1. It can
+#be changed to another value through this parameter for OSP17.
+osds_per_device: 1
 #osd_pool_default_pg_num:
 #osd_pool_default_pgp_num:
 
--- a/templates/internal.yml.j2
+++ b/templates/internal.yml.j2
@@ -14,10 +14,12 @@ parameter_defaults:
   CephPoolDefaultPgNum: 32
   CephAnsiblePlaybookVerbosity: 1
   CephAnsibleDisksConfig:
+{% if osp_release|int < 17 %}
     devices:
 {% for disk in storage_node_disks[1:] %}
       - {{ disk }}
 {% endfor %}
+{% endif %}
 
 # the following two parameters are the defaults. Just included them here for info
     osd_scenario: {{ osd_scenario }}


post_introspect.yml
```diff
@@ -25,8 +25,7 @@
             openstack baremetal introspection data save {{ item }} | jq . | grep by-path | cut -d \" -f 4
           register: storage_disks_info
           with_items: "{{ total_nodes.stdout_lines | default([]) }}"
-          when: composable_roles != true
-
+          when: composable_roles != true and osp_release|int < 17
```
rajeshP524: Nit:
I don't think 'osp_release|int < 17' is required here. because this task is defined under a block, which already satisfies this condition[1].
[1] https://github.com/redhat-performance/jetpack/pull/497/commits/888385e4a6b05afd25f48a7900f1be8c12d351ef#diff-e9917bb078e437a22da6c25a917233494016d5834226d93d68a250488d23c790R55

templates/osd-spec.yml.j2
```diff
@@ -0,0 +1,10 @@
+data_devices:
+{%if storage_node_disks is defined %}
+  paths:
+{% for disk in storage_node_disks %}
+    - '{{disk}}'
+{% endfor %}
+{% else %}
+  all: true
+{% endif %}
+osds_per_device: {{ osds_per_device }}
```
rajeshP524, line 10: From where this particular var/fact(osds_per_device) is coming from ?

rajeshP524:
I think.. this task[1] has to be executed only for 'osp_release|int < 17'. Because,
-  we are not collecting any disks information from introspection data for OSP17 with this PR now. Therefore a 'storage_node_disks' fact is not required for OSP17 and the only place it comes from is group_vars/all.yml.
- And when 'storage_node_disks' is not defined in group_vars/all.yml.. its handled here anyways[2]

[1] https://github.com/redhat-performance/jetpack/blob/master/post_introspect.yml#L119-L122
[2] https://github.com/redhat-performance/jetpack/commit/888385e4a6b05afd25f48a7900f1be8c12d351ef#diff-fd69b572ea70974f394344c9e9b5f346ef4be0ea23c5b934013e50c5ccf7f485R7-R8

post_introspect.yml
kemmy22: yea, right. This isn't needed.

templates/osd-spec.yml.j2
kemmy22: it is supposed to come from group vars , which i missed adding :)
let me get that line back

kemmy22:
spot on reg setting that fact, it looks correct to me what you just said. I'll still re verify once.

kemmy22:
I just checked that, and I had that check in place from the test directory. Missed out on inducting it in the commit.

rajeshP524:
LGTM!

overcloud.yml
```diff
@@ -281,3 +281,12 @@
...
             chdir: "{{ infrared_dir }}"
+        when: (osp_release|int < 17) or (osp_release|int == 17 and not ceph_enabled)
+
+      - name: run tripleo-overcloud deploy with ceph for osp 17
+        shell: |
+            source .venv/bin/activate
+            infrared tripleo-overcloud -vvv --version {{ osp_release }} --deployment-timeout {{ deployment_timeout | default(240) }} --build {{ osp_puddle }} --deployment-files {{ nic_configs }} --introspect no --tagging no --deploy yes --controller-nodes {{ controller_count }} --compute-nodes {{ compute_count }} --storage-nodes {{ ceph_node_count }} {{ oc_extra_templates }} --network-protocol ipv4 --network-backend {{ network_backend }} {{ network_type }} true {{ ceph_params | default('') }} {{ storage_protocol_backend | default('') }} --ceph-osd-spec-file {{ osd_spec_file }} --ntp-server {{ gateway }} --public-network false {{ oc_heat_configs | default('') }} {{ oc_config_resource | default('') }} > {{ log_directory }}/overcloud_deploy.log 2>&1
+        args:
+            chdir: "{{ infrared_dir }}"
+        when: osp_release|int == 17 and ceph_enabled
```
sanjaychari: Should this be changed to `osp_release|int >= 17` for now? If there are deployment changes in future releases, we can change it again later. I think it might be better to assume that there will not be deployment changes in this aspect in future releases, and in that case, we wouldn't have to change the condition check again if it is `osp_release|int >= 17`. Thoughts?

templates/osd-spec.yml.j2
```diff
@@ -0,0 +1,10 @@
+data_devices:
```
sanjaychari, line 1: Doesn't cephadm detect all storage disks except the root disk automatically and use them for deployment? If yes, why do we need to pass this list from the templates?

overcloud.yml
kemmy22: As discussed, we should be changing all the conditions where we have used == 17 to >= 17, so i'll make this change.

post_introspect.yml
```diff
@@ -119,11 +118,23 @@
...
           force: yes
-        when: ceph_enabled
+        when: ceph_enabled and osp_release|int < 17
+
+      - name: generate osd-spec.yml.j2
+        template:
+          src: osd-spec.yml.j2
+          dest: "{{ ansible_user_dir }}/osd-spec.yml"
+          force: yes
+        when: ceph_enabled and osp_release|int == 17
```
sanjaychari: I am still seeing `==17` in some places. Can you change it to `>=17` wherever it is `==17` in this PR?

overcloud.yml
```diff
@@ -281,3 +281,12 @@
             infrared tripleo-overcloud -vvv --version {{ osp_release }} --deployment-timeout {{ deployment_timeout | default(240) }} --build {{ osp_puddle }} --deployment-files {{ nic_configs }} --introspect no --tagging no --deploy yes --controller-nodes {{ controller_count }} --compute-nodes {{ compute_count }} --storage-nodes {{ ceph_node_count }} --overcloud-image-name full {{ oc_extra_templates }} --network-protocol ipv4 --network-backend {{ network_backend }} {{ network_type }} true {{ ceph_params | default('') }} {{ storage_protocol_backend | default('') }} --ntp-server {{ gateway }} --public-network false {{ oc_heat_configs | default('') }} {{ oc_config_resource | default('') }} > {{ log_directory }}/overcloud_deploy.log 2>&1
         args:
             chdir: "{{ infrared_dir }}"
+        when: (osp_release|int < 17) or (osp_release|int == 17 and not ceph_enabled)
```
sanjaychari: one last place to change to >=17 :)

kemmy22: lol

sanjaychari (approved):
LGTM

post_introspect.yml
```diff
@@ -119,11 +118,23 @@
...
+        when: ceph_enabled and osp_release|int < 17
+
+      - name: generate osd-spec.yml.j2
+        template:
+          src: osd-spec.yml.j2
+          dest: "{{ ansible_user_dir }}/osd-spec.yml"
+          force: yes
+        when: ceph_enabled and osp_release|int >= 17
+
+      - name: set fact for osd-spec.yml file
```
masco: move this task above and use this osd_spec_file fact at line no.133 and both the task can be added in single block

kemmy22: done

kemmy22:
Added a condition to specify ceph_params based on osp_release, merging it into a single deployment command.
Test case passed, deployment in progress.

overcloud.yml
```diff
@@ -281,3 +281,12 @@
...
         args:
             chdir: "{{ infrared_dir }}"
+        when: (osp_release|int < 17) or (osp_release|int >= 17 and not ceph_enabled)
+
+      - name: run tripleo-overcloud deploy with ceph for osp 17
+        shell: |
+            source .venv/bin/activate
+            infrared tripleo-overcloud -vvv --version {{ osp_release }} --deployment-timeout {{ deployment_timeout | default(240) }} --build {{ osp_puddle }} --deployment-files {{ nic_configs }} --introspect no --tagging no --deploy yes --controller-nodes {{ controller_count }} --compute-nodes {{ compute_count }} --storage-nodes {{ ceph_node_count }} {{ oc_extra_templates }} --network-protocol ipv4 --network-backend {{ network_backend }} {{ network_type }} true {{ ceph_params | default('') }} {{ storage_protocol_backend | default('') }} --ceph-osd-spec-file {{ osd_spec_file }} --ntp-server {{ gateway }} --public-network false {{ oc_heat_configs | default('') }} {{ oc_config_resource | default('') }} > {{ log_directory }}/overcloud_deploy.log 2>&1
```
asyedham: use set_fact to set the value of osd_spec_file instead of having two separate tasks to run tripleo-overcloud deploy

asyedham (changes requested):

overcloud.yml
kemmy22: i've included that change already, will push the change in a while.

kemmy22:
Added suggested changes and moved some piece of code to overcloud.yml

overcloud.yml
```diff
@@ -257,14 +257,26 @@
...
 
       - name: set deployment_timeout to 2400
         set_fact:
           deployment_timeout: 2400
         when: scale_compute_vms == true
 
+      - block:
+        - name: set fact for osd-spec.yml file
+          set_fact:
+            osd_spec_file: "{{ ansible_user_dir }}/osd-spec.yml"
```
asyedham, line 271: shouldn't  osd_spec_file fact set before `set facts for ceph deployment` task as it will fail otherwise
https://github.com/redhat-performance/jetpack/pull/497/files#diff-962dc7050a566bd835f3065a4da65d205c5ca498df81ec4262079970ee417fd7R258-R260

asyedham (changes requested):

overcloud.yml
asyedham: ignore my comment before as the value is string for ceph_params

post_introspect.yml
```diff
@@ -53,12 +52,12 @@
             storage_node_disks: "{{ storage_node_disks|default([]) + item.stdout_lines }}"
           with_items: "{{ storage_disks_info_comp.results }}"
           when: composable_roles
-      when: storage_node_disks is not defined and ceph_enabled
+      when: storage_node_disks is not defined and ceph_enabled and osp_release|int < 17
```
asyedham, line 55: Currently jetpack supports to 2 options for setting storage disks for ceph based deployments.
1. In group_vars/all.yml  explicitly specify the storage disks
https://github.com/redhat-performance/jetpack/blob/master/group_vars/all.yml#L281
2.  use the introspection data to identify the disks 
Why is the PR removing support for the 2nd option?

asyedham (changes requested):

post_introspect.yml
kemmy22: With Ceph 5, Ceph can automatically created OSDs on the disks available on the ceph host which wasn't an option available until OSP16 which is why we were using facts from introspection data.
Although, if we still want specific disks we can specify that through storage_node_disks in group vars or we need not set facts using introspection data during the time of deployment as Ceph will identify all of the disks available on the storage hosts and create OSDs on it which will form the cluster.
With this approach, by utilising ceph we also save good amount of time during deployment.

asyedham (approved):
LGTM!

overcloud.yml
kemmy22: okay, i can still move it up if you suggest 👍

post_introspect.yml
asyedham: ack, but in that case PR is only supporting when storage_disks is specified but what about if it wants to create OSDs on all the available disks?

kemmy22: Yes, in that case osd-spec.yml.j2 will default to else condition which says:
https://github.com/redhat-performance/jetpack/pull/497/commits/263985fb16883e673351293dc8ac3427fa092587#diff-fd69b572ea70974f394344c9e9b5f346ef4be0ea23c5b934013e50c5ccf7f485R8

And it will then create OSDs on all devices available on that host.

PR #498, opened 2022-10-31
Removes the changes for customising overcloud partitions.

kemmy22:
With overcloud-full image theses changes are no longer needed. 
Although the overcloud-node-growvols.yml file is left intact should we need to use it at any point in time later.


rajeshP524:
LGTM!

Issue #433, opened 2022-02-23
Avoid using ovs bridges for controlplane traffic


links84:
ovs bridges are only required for tenant network or provider network. If the external api network is also used for instance floating-ip or provider network then that would also need ovs bridges.
networks like storage, internal-api or ctlplane is not recommended to be kept under ovs to reduce load on ovs.

PR #501, opened 2022-11-09
Minimise use of ovs bridges in OSP17

sanjaychari:
This PR moves internal-api, storage and storage-mgmt networks away from ovs bridges and onto the linux interface directly, to reduce load on ovs. This commit is for OSP 17. The changes for other OSP releases will be addressed in separate PRs. This PR has been tested on a 16 node environment with ceph enabled with composable roles for heterogeneous hardware.

Partially addresses #433

### DIFF
--- a/templates/ceph-storage.j2.j2
+++ b/templates/ceph-storage.j2.j2
@@ -14,22 +14,17 @@ network_config:
     default: true
 {% endraw %}
 
-- type: ovs_bridge
-  name: br-isolated
-  use_dhcp: false
-  members:
-  - type: interface
-    name: {{ storage_interface }}
-    primary: true
-  - type: vlan
-    vlan_id: {{ storage_network_vlan_id }}
-    {% raw -%}
-    addresses:
-    - ip_netmask: {{ storage_ip }}/{{ storage_cidr }}
+- type: vlan
+  device: {{ storage_interface }}
+  vlan_id: {{ storage_network_vlan_id }}
+  {% raw -%}
+  addresses:
+  - ip_netmask: {{ storage_ip }}/{{ storage_cidr }}
 {% endraw  %}
-  - type: vlan
-    vlan_id: {{ storage_mgmt_network_vlan_id }}
-    {% raw -%}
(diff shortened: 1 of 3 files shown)


pradiptapks (changes requested):
@sanjaychari It makes sense to remove the isolated network from the ovs-bridge. You can also remove `ctlplane` network mapping from ovs as well.

sanjaychari:
@pradiptapks ctlplane network is already not on ovs bridge in OSP 17 when external network is not on the same interface as ctlplane network. 
https://github.com/redhat-performance/jetpack/blob/master/templates/controller.j2.j2#L29-L44

When external network and ctlplane network have to be on same interface due to lack of sufficient interfaces, then an ovs bridge is created to host both of them because external network needs ovs bridge.
https://github.com/redhat-performance/jetpack/blob/master/templates/controller.j2.j2#L67-L86

So I do not think any change is required for ctlplane network in OSP 17 templates. In OSP 16 templates, ctlplane network is using ovs-bridge br-ctlplane like you mentioned. So that will have to be changed in OSP 16. I will address the OSP 16 changes in a separate PR.

pradiptapks:
@sanjaychari Thank you for the details. You can create another PR for OSP16 network changes similar to OSP17 network configuration.

These chages can help for nfv best practices as well.

pradiptapks (approved):

pradiptapks:
@sanjaychari I don't have a environment to test pr. The chages looks good to me. But I prefer a 2nd thought and test from @masco & @asyedham.

templates/ceph-storage.j2.j2
```diff
@@ -14,6 +14,7 @@ network_config:
     default: true
 {% endraw %}
 
+{% if manila_enabled %}
```
asyedham: Please submit a separate PR for manila related changes as jetpack is not yet supporting it for OSP17

asyedham (changes requested):

asyedham (approved):
LGTM!

PR #502, opened 2022-11-09
fix composable roles for osp < 17

asyedham:
- This PR defines flavor that needs to be used for controller
   when controller_machine_type is defined
- adjusts the overcloud deploy script


masco (approved):
LGTM

sanjaychari (approved):
LGTM

PR #505, opened 2022-11-15
automate the ceph_ifaces and controller_ifaces

asyedham:
This PR removes the dependency of setting ceph_ifaces and controller_ifaces in group_vars/all.yml


sanjaychari:
This is not specifically related to this PR, but we should probably either maintain or deprecate all.yml files in https://github.com/redhat-performance/jetpack/tree/master/ci, as they are out-of-date.

sanjaychari (approved):

masco:
@asyedham i am seeing some .swp files in the list, can you check it

composable_prepare_nic_configs.yml
```diff
@@ -230,10 +234,10 @@
         when: osp_release|int < 17
 
       - block:
-          - name: fail if ceph_machine_type or ceph_ifaces are not defined
+          - name: fail if ceph_machine_type
```
masco: name not giving any meaning

masco: I am thinking of moving this type of check at the beginning of the installation, instead of doing it halfway. thoughts?

asyedham: Done

common.yml
```diff
@@ -1,6 +1,11 @@
 ---
 - hosts: localhost
   tasks:
+      - name: fail if ceph_machine_type
```
masco: job name can be, "fail if ceph_machine_type is not defined"

overcloud.yml
```diff
@@ -265,12 +265,16 @@
             src: osd-spec.yml.j2
             dest: "{{ osd_spec_file }}"
             force: yes
+
+        - name: set facts for ceph deployment
+          set_fact:
+            ceph_params: "{{ ('--storage-backend ceph --storage-external no --ceph-osd-spec-file ' +  osd_spec_file) }}"
```
masco: ceph_params: "--storage-backend ceph --storage-external no --ceph-osd-spec-file {{ osd_spec_file }}"
can you check this will work?

asyedham: Yes

masco (approved):
LGTM

PR #507, opened 2022-11-16
fix creating the flavor

asyedham:
This PR fixes the corner case when total count of
machine type is equal to controller machine type
count.


masco (approved):
LGTM

PR #511, opened 2022-11-24
Ignore any failures while cleaning network ifaces

### LINKED ISSUE #510: Ignore any failures while cleaning network interfaces during undercloud setup stage
When we provision undercloud using a different image (other than foreman) and deploy using jetpack the deployment fails at prepare undercloud stage due to the absence of clean-interfaces.sh script to remove the existing network config in case of an image from foreman.
However, if we're using another image that config won't be there anyhow which leads to failure in deployment.

kemmy22:
Ignore any failures while cleaning network ifaces on undercloud, during undercloud prepare stage.
This PR resolves https://github.com/redhat-performance/jetpack/issues/510

### DIFF
--- a/prepare_undercloud.yml
+++ b/prepare_undercloud.yml
@@ -21,6 +21,7 @@
       shell: |
         /root/clean-interfaces.sh --nuke
       changed_when: false
+      ignore_errors: true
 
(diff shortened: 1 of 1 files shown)


sanjaychari (approved):
LGTM

PR #512, opened 2022-11-24
Enables Manila for OSP17

### LINKED ISSUE #508: Add support for Manila in OSP17
Enable manila

kemmy22:
Enables support for manila with nfsganesha for OSP17.
Also corrects tenant network mapping in network-env file. It was earlier pointing to br-isolated birdge whereas tenant network in the nic config file is attached to br-tenant bridge.
This PR resolves https://github.com/redhat-performance/jetpack/issues/508

### DIFF
--- a/composable.yml
+++ b/composable.yml
@@ -113,3 +113,14 @@
         openstack overcloud roles generate -o ~/roles_data.yaml --roles-path ~/roles {{ roles }}
       args:
         chdir: "/home/stack"
+
+    - block:
+      - name: create roles directory as required for infrared
+        file:
+          path: "~/virt/roles"
+          state: directory
+
+      - name: copy roles_data file to virt directory
+        shell: |
+          cp ~/roles_data.yaml ~/virt/roles/
+      when: ceph_enabled and osp_release|int >= 17
--- a/templates/network-environment_v2.yaml.j2
+++ b/templates/network-environment_v2.yaml.j2
@@ -1,7 +1,11 @@
 # v2 config
 
 resource_registry:
+{% if manila_enabled %}
+    OS::TripleO::ControllerStorageNfs::Net::SoftwareConfig: OS::Heat::None
+{% else %}
     OS::TripleO::Controller::Net::SoftwareConfig: OS::Heat::None
+{% endif %}
 {% if composable_roles %}
 {% for node_type in machine_types %}
     OS::TripleO::Compute{{ node_type }}::Net::SoftwareConfig: OS::Heat::None
@@ -13,7 +17,11 @@ resource_registry:
     OS::TripleO::CephStorage::Net::SoftwareConfig: OS::Heat::None
 {% endif %}
 parameter_defaults:
+{% if manila_enabled %}
+    ControllerStorageNfsNetworkConfigTemplate: '/home/stack/virt/network/vlans/controller.j2'
+{% else %}
     ControllerNetworkConfigTemplate: '/home/stack/virt/network/vlans/controller.j2'
+{% endif %}
 {% if composable_roles %}
 {% for node_type in machine_types %}
     Compute{{ node_type }}NetworkConfigTemplate: '/home/stack/virt/network/vlans/compute_{{ node_type }}.j2'
@@ -24,8 +32,13 @@ parameter_defaults:
 {% if ceph_enabled %}
     CephStorageNetworkConfigTemplate: '/home/stack/virt/network/vlans/ceph-storage.j2'
 {% endif %}
+{% if manila_enabled %}
+    ControllerStorageNfsParameters:
+      NeutronBridgeMappings: "datacentre:br-ex,tenant:br-tenant,storage:br-isolated"
+{% else %}
     ControllerParameters:
-      NeutronBridgeMappings: "datacentre:br-ex,tenant:br-isolated"
+      NeutronBridgeMappings: "datacentre:br-ex,tenant:br-tenant"
+{% endif %}
 {% if composable_roles %}
 {% if dvr_enabled == false %}
 {% for node_type in machine_types %}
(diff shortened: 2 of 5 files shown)


templates/network-environment_v2.yaml.j2
```diff
@@ -24,13 +24,22 @@ parameter_defaults:
 {% if ceph_enabled %}
     CephStorageNetworkConfigTemplate: '/home/stack/virt/network/vlans/ceph-storage.j2'
 {% endif %}
+{% if manila_enabled %}
     ControllerParameters:
```
asyedham: this should be ControllerStorageNfsParameters:

composable.yml
```diff
@@ -8,6 +8,14 @@
       shell: |
         cp   {{ oc_deploy_template }} {{ infrared_oc_deploy_template }}
 
+    - name: override the overcloud_ceph_deploy.sh.j2 for composable roles
+      vars:
+        oc_ceph_deploy_template: "{{ ansible_user_dir }}/jetpack/templates/composable_roles_overcloud_ceph_deploy.sh.j2"
+        infrared_oc_ceph_deploy_template:  "{{ ansible_user_dir }}/.infrared/plugins/tripleo-overcloud/templates/overcloud_ceph_deploy.sh.j2"
+      shell: |
+        cp   {{ oc_ceph_deploy_template }} {{ infrared_oc_ceph_deploy_template }}
+      when: osp_release|int >= 17
```
asyedham: add check as this task has to be executed only when ceph enabled

asyedham (changes requested):

templates/network-environment_v2.yaml.j2
asyedham: same here as well add checks for ControllerStorageNfs
https://github.com/redhat-performance/jetpack/blob/d9fc1892885d421619f9285bda99a44651d73bdc/templates/network-environment_v2.yaml.j2#L16
https://github.com/redhat-performance/jetpack/blob/d9fc1892885d421619f9285bda99a44651d73bdc/templates/network-environment_v2.yaml.j2#L4

kemmy22: yes i've added that too with this change, deployment in progress.

composable.yml
kemmy22: added the changes

templates/network-environment_v2.yaml.j2
kemmy22: added the changes

asyedham: this is still missing can you please add it

```
{% if manila_enabled %}
       OS::TripleO::ControllerStorageNfs::Net::SoftwareConfig: OS::Heat::None
{% else %}
       OS::TripleO::Controller::Net::SoftwareConfig: OS::Heat::None 
{% endif %}
```

kemmy22: thanks for pointing

templates/composable_roles_overcloud_ceph_deploy.sh.j2
```diff
@@ -0,0 +1,13 @@
...
+openstack overcloud ceph deploy -v \
+-o /home/stack/templates/overcloud-ceph-deployed.yaml \
+--container-image-prepare "{{ containers_home_dir }}/containers-prepare-parameter.yaml" \
+{% if install.ceph.initial.conf.file %} --config {{ ansible_user_dir }}/initial-ceph.conf{% endif %} \
+{% if install.ceph.osd.spec.file %} --osd-spec {{ ansible_user_dir }}/osd-spec.yml{% endif %} \
+{% if ceph_node_cnt|int < 3 %}--single-host-defaults {% endif %} \
+--stack "{{ install.overcloud.stack }}" \
+{% if install.ceph.cluster.name and install.ceph.cluster.name|length > 0 %}--cluster {{ install.ceph.cluster.name }}{% endif %} \
+--network-data "{{ template_base }}/network/network_data_v2.yaml" \
+--roles-data "/home/stack/roles_data.yaml" \
```
asyedham: can you make the necessary changes in the roles_data_path in the code instead of maintaining a duplicate of [overcloud_ceph_deploy.sh.j2](https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/templates/overcloud_ceph_deploy.sh.j2)

asyedham (changes requested):

templates/composable_roles_overcloud_ceph_deploy.sh.j2
kemmy22: We already generate roles_data on the undercloud, so the idea was to to just use it from the undercloud node like it is being done for composable roles currently.
The other way around, we'll have to copy the generated roles_data file from undercloud to jumphost and then override infrared. This seems to be a simple & similar approach to what we are already doing, hence went with this.

asyedham: If you update --roles-path  to ~/virt/roles (as it is in infrared) and other references of roles path in the jetpack code that should be sufficient no need to maintain a duplicate of [overcloud_ceph_deploy.sh.j2](https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/templates/overcloud_ceph_deploy.sh.j2).  Let me know if still have any doubts

kemmy22: Made the amendments

composable.yml
```diff
@@ -113,3 +113,14 @@
         openstack overcloud roles generate -o ~/roles_data.yaml --roles-path ~/roles {{ roles }}
       args:
         chdir: "/home/stack"
+
+    - block:
+      - name: create roles directory as required for infrared
```
asyedham, line 118: @kemmy22 instead of creating separate check for OSP 17, can you please update the roles-path to ~/virt/roles to all the possible occurrences in jetpack code that  will avoid additional checks

asyedham (changes requested):

composable.yml
kemmy22: @asyedham I think we can track that through a separate issue/change. What do you think? Since this change is for ceph & manila.

asyedham: ack

kemmy22:
Rebased.

asyedham (approved):
LGTM!

PR #516, opened 2022-11-28
Add support for OSP 17.1

### LINKED ISSUE #513: OSP 17.1 deployment failing because of missing osp_rhel_mapping
There is currently no osp_rhel_mapping for OSP 17.1.
https://github.com/redhat-performance/jetpack/blob/b48d6bbe632a4f774605467c449ca2f68928f7cf/setup_undercloud.yml#L5-L13

This causes the following error.
```
TASK [set os_search_string] **********************************************************************************************************************
task path: /home/sanjay/jetpack/tasks/install_os.yml:2
Thursday 24 November 2022  07:41:05 -0500 (0:00:00.246)       0:07:34.594 ***** 
fatal: [localhost]: FAILED! => {
    "msg": "The task includes an option with an undefined variable. The error was: RHEL {{ osp_rhel_mapping[osp_release|float] }}: dict object has
 no element 17.1\n\nThe error appears to be in '/home/sanjay/jetpack/tasks/install_os.yml': line 2, column 3, but may\nbe elsewhere in the file de
pending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: set os_search_string\n  ^ here\n"
}
```

OSP 17.1 GA will use RHEL 9.2. We should set osp_rhel_mapping to 9.1 for now for OSP 17.1, and later change it to 9.2.

### LINKED ISSUE #514: OSP 17.1: Issue in task "update host to install new os"
RHEL 9.1 is not available in the scale lab Foreman instance yet. Because of this, the following error occurs while using RHEL 9.1 for OSP 17.1.
 
```
 "cmd": "podman run --rm hammercli hammer -u cloud11 -p rdu2@2351  host update --name f24-h01-000-r630.rdu2.scalelab.redhat.com --build 1 --ope
ratingsystem 'RHEL 9.1' --medium 'RHEL Local' --partition-table 'RHEL9-ONLY' --overwrite 1\n", 
    "delta": "0:00:02.186455", 
    "end": "2022-11-24 08:04:23.232023", 
    "invocation": {
        "module_args": {
            "_raw_params": "podman run --rm hammercli hammer -u cloud11 -p rdu2@2351  host update --name f24-h01-000-r630.rdu2.scalelab.redhat.co$
 --build 1 --operatingsystem 'RHEL 9.1' --medium 'RHEL Local' --partition-table 'RHEL9-ONLY' --overwrite 1\n", 
            "_uses_shell": true, 
            "argv": null, 
            "chdir": null, 
            "creates": null, 
            "executable": null, 
            "removes": null, 
            "stdin": null, 
            "stdin_add_newline": true, 
            "strip_empty_ends": true, 
            "warn": true
        }
    }, 
    "msg": "non-zero return code", 
    "rc": 70, 
    "start": "2022-11-24 08:04:21.045568", 
    "stderr": "Could not update the host:\n  Error: operatingsystem not found.", 
    "stderr_lines": [
        "Could not update the host:", 
        "  Error: operatingsystem not found."
    ],
```

We probably need to change some condition checks like https://github.com/redhat-performance/jetpack/blob/c7746540b7defdc0ce810bde702e5e13415bd6a1/tasks/install_os.yml#L63 until RHEL 9.1 is available in the scale lab Foreman instance.

sanjaychari:
This PR makes changes required for OSP 17.1.
The latest passed_phase2 puddle of OSP 17.1 requires RHEL 9.1.
RHEL 9.1 is not yet available in the Scale Lab Foreman, so the
upgrade script provided by the lab team is used to upgrade to
RHEL 9.1 after initially booting RHEL 9.0.

Resolves #513 
Resolves #514

### DIFF
--- a/setup_undercloud.yml
+++ b/setup_undercloud.yml
@@ -11,6 +11,7 @@
         16.1: 8.2
         16.2: 8.4
         17.0: 9.0
+        17.1: 9.1
   tasks:
     - name: set async_install to empty list
       set_fact:
--- a/tasks/install_os.yml
+++ b/tasks/install_os.yml
@@ -16,19 +16,28 @@
   set_fact:
     rhel_stdout: "{{ (rhel_release.results[0].stdout) | ternary(rhel_release.results[0].stdout, rhel_release.results[1].stdout) }}"
 
-# forman is not having 8.4. So we need to install previsous version and upgrade later
-# We don't install RHEL 8.2 or RHEL 8.4 when host already has this for OSP16.2
-- name: set os_install when installed OS version not matching with the needed
+# Foreman does not have RHEL 8.4. So we need to install the previous version and upgrade later.
+# We don't install RHEL 8.2 or RHEL 8.4 when the host already has this for OSP16.2.
+- name: set os_install when installed OS version not matching with the needed (RHEL < 9.0)
   vars:
     lab_available_os: "{{ (needed_os == 'RHEL 8.4') | ternary('RHEL 8.2', needed_os) }}"
   set_fact:
     os_install: "{{ lab_available_os }}"
-  when: (force_reprovision == true) or (rhel_stdout == "" or ((needed_os.split()[1] not in rhel_stdout) and (lab_available_os.split()[1] not in rhel_stdout)) or (os_search_string not in rhel_stdout))
+  when: ((force_reprovision == true) or (rhel_stdout == "" or ((needed_os.split()[1] not in rhel_stdout) and (lab_available_os.split()[1] not in rhel_stdout)) or (os_search_string not in rhel_stdout))) and (needed_os.split()[1] is version('9.0', '<'))
+
+# Foreman does not have RHEL 9.1. So we need to install the previous version and upgrade later.
+# We don't install RHEL 9.0 or RHEL 9.1 when the host already has this for OSP17.1.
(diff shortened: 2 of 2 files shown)


setup_undercloud.yml
```diff
@@ -22,6 +23,7 @@
         hypervisor_host: "{{ undercloud_hostname }}"
         hypervisor_password: "{{ ansible_ssh_pass }}"
       include_tasks: tasks/install_os.yml
+      when: osp_release is version('17.1', '<')
```
masco: This need to be removed once the 9.1 image is available right? if so, add a TODO comment

sanjaychari: @masco I think RHEL 9.1 is not yet the default in passed_phase2 for OSP 17.1.
@asyedham What do you think? Should I change to 9.0? In that case we don't need to skip this now.

asyedham: Yeah we can do that @masco what are your thoughts?

sanjaychari: The latest passed_phase2 puddle of OSP 17.1 uses RHEL 9.1.

masco (approved):
LGTM

asyedham:
Just a NIT, can you change the commit message to `Add support for OSP 17.1`

sanjaychari:
> Just a NIT, can you change the commit message to `Add support for OSP 17.1`

Done

PR #519, opened 2022-12-07
fix mount nvme in compute nodes

masco:


asyedham (approved):
LGTM!

overcloud.yml
```diff
@@ -27,10 +27,16 @@
...
-        dest: "/home/stack/firstboot-nvme.yaml"
+    - block:
+        - name: generate firstboot-nvme.yaml
+          template:
+            src: "firstboot-nvme.yaml.j2"
+            dest: "/home/stack/firstboot-nvme.yaml"
+        - name: override the firstboot.yaml
+          template:
+            src: "firstboot.yaml.j2.j2"
+            dest: "{{ hostvars['localhost'].ansible_user_dir }}/.infrared/plugins/tripleo-overcloud/templates/firstboot.yaml.j2"
```
sanjaychari, line 38: If this is delegated to localhost, is the hostvars still required?

masco: yes Sanjay, it was picking the remote machine value so I used the hostvars

PR #153, opened 2020-02-10
Add tls-everywhere

asyedham:

PR #520, opened 2022-12-08
Add support for TLS-Everywhere

sanjaychari:
This commit adds support for TLS-Everywhere to Jetpack. It has been tested on OSP 17.

### DIFF
--- a/browbeat.yml
+++ b/browbeat.yml
@@ -24,6 +24,7 @@
           chdir: "{{ browbeat_dir }}/ansible"
     - name: install browbeat
       shell: |
+          source /home/stack/.bash_profile
           ansible-playbook -i hosts.yml install/browbeat.yml
       args:
           chdir: "{{ browbeat_dir }}/ansible"
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -296,3 +296,11 @@ manila_enabled: false
 # DVR is enabled by default for osp_releases >= 16, set this to false for non-dvr based deployments.
 # This variable has no affect on osp_releases < 16, as jetpack doesn't support dvr for osp_releases < 16 as of now.
 dvr_enabled: true
+
+# TLS-Everywhere has been tested on OSP 17.
+tls_everywhere: false
+# vars for TLS-Everywhere deployment
+idm_host_fqdn:
+ipa_domain: redhat.local
+ipa_realm: REDHAT.LOCAL
+ipa_password: redhatidm
--- a/install_idm_server.yml
+++ b/install_idm_server.yml
@@ -1,0 +1,96 @@
+- hosts: localhost
+  vars:
+    chassis_password:  "{{ instackenv_content.nodes[0].pm_password }}"
+    idm_osp_rhel_mapping:
+      17.0: 9.0
+      17.1: 9.1
+  tasks:
+    - name: add IdM host to inventory file
+      add_host:
+        name: "{{ idm_host_fqdn }}"
+        groups: "idm"
+        ansible_host: "{{ idm_host_fqdn }}"
+        ansible_user: "root"
+        ansible_ssh_private_key_file: "{{ ansible_ssh_key }}"
+        ansible_python_interpreter: "{{ python_interpreter }}"
+
+    - name: install os on IdM node
+      vars:
+        needed_os: "RHEL {{ idm_osp_rhel_mapping[osp_release|float] }}"
+        hypervisor_host: "{{ idm_host_fqdn }}"
+        hypervisor_password: "{{ ansible_ssh_pass }}"
+      include_tasks: tasks/install_os.yml
+
+    - name: remove entry for IdM host from known_hosts file
+      known_hosts:
+        path: "{{ ansible_user_dir }}/.ssh/known_hosts"
+        name: "{{ idm_host_fqdn }}"
+        state: absent
+
+- hosts: idm
+  become: yes
+  become_user: root
+
+  tasks:
+    - name: Enable FIPS
+      shell: |
+        fips-mode-setup --enable
+
+    - name: Reboot the host
+      reboot:
+        reboot_timeout: 1200
+
+    - name: Get IP address of host
+      shell: |
+        ping {{ idm_host_fqdn }} -c 1 | head -1 | cut -d " " -f 3 | tr -d "(" | tr -d ")"
+      register: idm_host_ip_address
+
+    - name: Add entry in /etc/hosts
+      lineinfile:
+        path: /etc/hosts
+        line: '{{ idm_host_ip_address.stdout }} {{ idm_host_fqdn }} idm.{{ ipa_domain }}'
+
+    - name: Provide access to ports required by the IdM server
+      shell: |
+        systemctl unmask firewalld.service
+        systemctl start firewalld.service
+        systemctl enable firewalld.service
+        firewall-cmd --permanent --add-port={80/tcp,443/tcp,389/tcp,636/tcp,88/tcp,88/udp,464/tcp,464/udp,53/tcp,53/udp,123/udp}
+        firewall-cmd --permanent --add-service={freeipa-ldap,freeipa-ldaps,dns}
+        firewall-cmd --reload
+
+    - name: Install epel-release
+      dnf:
+        name: 'https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm'
+        state: present
+
+    - name: Install repositories for IdM
+      dnf:
+        name:
+          - ipa-server
+          - ipa-server-dns
+        state: present
+
+    - name: Get umask
+      shell: |
+        umask
+      register: current_umask
+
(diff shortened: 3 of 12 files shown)


browbeat.yml
```diff
@@ -27,6 +27,14 @@
           ansible-playbook -i hosts.yml install/browbeat.yml
       args:
           chdir: "{{ browbeat_dir }}/ansible"
+      when: tls_everywhere != true
+    - name: install browbeat
+      shell: |
+          source /home/stack/.bash_profile
```
asyedham, line 27: why do we need to separate tasks for installing browbeat?

group_vars/all.yml
```diff
@@ -290,3 +290,16 @@ osds_per_device: 1
...
+tls_everywhere: false
+# vars for TLS-Everywhere deployment
+# The host specified as idm_host_fqdn must not be present
+# in instackenv.json, and must be in the same allocation
+# that the OSP deployment will be in.
+idm_host_fqdn:
+ipa_domain:
+ipa_realm:
+ipa_password:
+ipa_ntp_server:
```
asyedham: can you minimize the variables to be set by users. Instead set default values

install_idm_server.yml
```diff
@@ -0,0 +1,146 @@
...
+    - name: add IdM host to inventory file
+      add_host:
+        name: "{{ idm_host_fqdn }}"
+        groups: "idm"
+        ansible_host: "{{ idm_host_fqdn }}"
+        ansible_user: "root"
+        ansible_ssh_private_key_file: "{{ ansible_ssh_key }}"
+        ansible_python_interpreter: "{{ python_interpreter }}"
+
+    - name: Install RHEL 8.2 on the IdM host
```
asyedham: any specific reason to use RHEL 8.2?

asyedham (changes requested):
Please automate removing IDM node from instackenv.json

group_vars/all.yml
asyedham: @sanjaychari you can do something like this for setting default values for IDM vars
https://github.com/redhat-performance/jetpack/pull/153/files#diff-7be6987d456ad921f461608b488a7593f037ba8fc6e43ed862c903e91109c00aR129-R139
https://github.com/redhat-performance/jetpack/pull/153/files#diff-0c5e8f9ce36621141ff343acf17436f40ed11c0e0e931b70b0ed28d1dabe6769R1-R40

undercloud.yml
```diff
@@ -39,6 +96,14 @@
         template:
             src: undercloud.conf.j2
             dest: "{{ undercloud_conf }}"
+
+      - name: Set overcloud_domain_name in undercloud.conf
+        lineinfile:
+          path: "{{ undercloud_conf }}"
```
asyedham: can you customize overcloud_domain_name in the template  undercloud.conf.j2 based on tls is enabled/disabled

asyedham (changes requested):

overcloud.yml
```diff
@@ -114,6 +114,48 @@
...
+        when: tls_everywhere == true
+
+      - name: set resource_registry tls-everywhere parameters
+        vars:
+          res_reg: "{{ resource_registry|default([]) }}"
+        set_fact:
+          resource_registry: "{{ res_reg + ['OS::TripleO::Services::IpaClient=/usr/share/openstack-tripleo-heat-templates/deployment/ipa/ipaservices-baremetal-ansible.yaml'] }}"
+        when: tls_everywhere == true
+
+      - name: add templates for tls-everywhere to extra_templates
```
asyedham, line 142: --tls-everywhere options not working, is infrared not adding this templates?

asyedham (changes requested):

undercloud.yml
```diff
@@ -1,3 +1,48 @@
+- hosts: undercloud
+  tasks:
```
asyedham, line 2: you can create separate tasks file for setting IDM server

asyedham (changes requested):

browbeat.yml
sanjaychari: `REQUESTS_CA_BUNDLE` is added to .bash_profile for the python requests module to recognize the IdM server as a valid certificate authority.
The .bash_profile file has to be sourced before Browbeat installation as it requires `REQUESTS_CA_BUNDLE` to be set as well.

group_vars/all.yml
sanjaychari: I can set default values for ipa_domain and ipa_realm.
But the user still has to set 4 variables : idm_host_fqdn(because they have to choose which host they want to use for the IdM server in their allocation), ipa_password, ipa_ntp_server and rhos_release_rpm_link since these are sensitive variables that cannot be set to default values.

install_idm_server.yml
sanjaychari: RHEL > 8.3 is required to enable FIPS(Federal Information Processing Standard). Since the packages come from the OSP 16.2 rhos-release puddle, RHEL 8.4 has to be installed on the IdM host. So we install RHEL 8.2 first and upgrade to RHEL 8.4.

undercloud.yml
sanjaychari: Ack

overcloud.yml
sanjaychari: I will try using this option.

undercloud.yml
sanjaychari: Ack

overcloud.yml
sanjaychari: infrared has some extra tasks with the --tls-everywhere flag that are not required in Jetpack.
This causes some errors during overcloud deployment.

```
TASK [tls-everywhere : fail] ***************************************************
task path: /home/sanjay/infrared/.venv/lib/python3.6/site-packages/infrared/common/roles/tls-everywhere/tasks/validate_overcloud.yml:10
fatal: [undercloud -> undercloud]: FAILED! => {
    "changed": false,
    "msg": "Network environment file with location /home/stack/virt/network/network-environment.yaml does not exist!"
}
```

browbeat.yml
asyedham: can we have one single task to install browbeat.

group_vars/all.yml
```diff
@@ -296,3 +296,13 @@ manila_enabled: false
...
 # This variable has no affect on osp_releases < 16, as jetpack doesn't support dvr for osp_releases < 16 as of now.
 dvr_enabled: true
+
+# TLS-Everywhere has been tested on OSP 17.
+tls_everywhere: false
+# vars for TLS-Everywhere deployment
+idm_host_fqdn:
+ipa_domain: redhat.local
+ipa_realm: REDHAT.LOCAL
+ipa_password:
```
asyedham: can we have a default password for ipa. If the user wants they can overwrite it

group_vars/all.yml
```diff
@@ -296,3 +296,13 @@ manila_enabled: false
...
+ipa_ntp_server:
```
asyedham: can we auto identify the ipa_ntp_server

asyedham (changes requested):

undercloud.yml
```diff
@@ -70,3 +89,16 @@
...
+      - name: Run the undercloud-ipa-install.yml playbook
+        shell: |
+          source /home/stack/.bash_profile
+          ansible-playbook --ssh-extra-args "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" /usr/share/ansible/tripleo-playbooks/undercloud-ipa-install.yaml
+        when: tls_everywhere == true
+        delegate_to: undercloud
+
+      - name: Update the undercloud for TLS-Everywhere
+        shell:
+          openstack undercloud install 2>&1 | tee undercloud_install.log
```
asyedham: can you add prefix `(time openstack undercloud install) &> undercloud_install_tls.log` as we want to capture the time taken to update undercloud

asyedham (changes requested):

group_vars/all.yml
```diff
@@ -296,3 +296,12 @@ manila_enabled: false
...
 dvr_enabled: true
+
+# TLS-Everywhere has been tested on OSP 17.
+tls_everywhere: false
+# vars for TLS-Everywhere deployment
+idm_host_fqdn:
+ipa_domain: redhat.local
+ipa_realm: REDHAT.LOCAL
+ipa_password: redhatidm
+rhos_release_rpm_link:
```
asyedham: can you check epel-release is sufficient to install packages

install_idm_server.yml
```diff
@@ -0,0 +1,146 @@
...
+      block:
+        - name: Build the images for hammercli
```
asyedham: can you please reuse [install_os.yml](https://github.com/redhat-performance/jetpack/blob/master/tasks/install_os.yml) for installing os on idm server

asyedham (changes requested):

asyedham (approved):
LGTM!

PR #521, opened 2022-12-08
fix the availability zone label on infra nodes

masco:
due to wrong avaiability zone, volume creation is failing this pr relabeling to the nova availability zone


venkataanil (approved):
LGTM

PR #525, opened 2022-12-21
set storage_node_disks manually for 1029u

asyedham:


group_vars/all.yml
```diff
@@ -277,21 +277,24 @@ ceph_machine_type: ''
 ceph_enabled: false
 osd_scenario: lvm
 osd_objectstore: bluestore
-#Note:By default storage_node_disks can be detected automatically
-#using introspection data
-#storage_node_disks: ['/dev/nvme0n1']
+#Note:By default storage_node_disks can be detected automatically using
```
sanjaychari, line 280: I recently used 1029utrtp nodes as ceph_machine_type even though the nodes were also being used for computes and controllers. I didn't face this issue. Did you face this issue during node provisioning?

asyedham: But it does detect the disk of other 1029utrp in storage_node_disks

sanjaychari: My question was, is it required to specify storage_node_disks manually for 1029u systems? It worked without that for me.

asyedham: Since we are getting 1029u with diff interface naming and in the group_vars/all.yml it is 1029utrtp/1029u10tnrt to distinguish it will not detect any disk as it tries to search for 1029utrtp/1029u10tnrt  in ipmi_addr [code](https://github.com/redhat-performance/jetpack/blob/master/post_introspect.yml#L34).Not sure how it worked for you. Also like mentioned in my reply above it would detect the disks from the other 1029u as well

masco (approved):

PR #529, opened 2022-12-23
Add support for scaleout with composable roles

rajeshP524:
This PR adds support for scaleout feature with and without composable roles.


scaleout.yml
```diff
@@ -0,0 +1,309 @@
...
+    - name: load instackenv
+      include_tasks: tasks/load_instackenv.yml
+
+    - name: get intrepeter
+      include_tasks: tasks/get_interpreter.yml
+      vars:
+        hostname: "{{ undercloud_hostname }}"
+        user: "stack"
+
+    - name: add undercloud
```
sanjaychari, line 5: If `cleanup.yml` has not been run after overcloud deployment, won't the undercloud still be present in the inventory from the previous deployment? In that case, it wouldn't have to be added to the inventory again.

Although, if a deployment on another allocation has been run before the scaleout, then this step would be required.

rajeshP524: scaleout.yml is independent of what allocation has been deployed before. 
It can be run on any allocation that is jetpack deployed. So its good to assume that inventory is absent.
Also I have made a change to pass "undercloud_host" via group_vars/all.yml, as undercloud host need not always be the first node in lab generated instackenv.json

scaleout.yml
```diff
@@ -0,0 +1,363 @@
...
+            - name: register the output of total new nodes
+              shell: |
+                cat /home/stack/total_nodes.txt
+              register: total_nodes
+
+        - name: set root hints
+          include_tasks: tasks/set_root_hints.yml
+          when: lab_name in ['scale', 'alias']
+
+    - name: create flavors
```
masco, line 211: NIT: check is it possible to use this https://docs.ansible.com/archive/ansible/2.5/modules/os_nova_flavor_module.html

masco (approved):
LGTM

PR #536, opened 2023-04-20
Add rhel9 interfaces for r750's

rajeshP524:


group_vars/all.yml
```diff
@@ -27,6 +27,8 @@ alias:
         rhel7_interfaces: [em2, em1, p7p1, p7p2]
         rhel8_interfaces: [eno2, eno1, ens7f0, ens7f1]
         rhel9_interfaces: [eno2, eno1, ens7f0, ens7f1]
+      r750:
```
sanjaychari, line 30: We usually add rhel7_interfaces and rhel8_interfaces also like this when adding a new section but the interfaces are not available.

rhel7_interfaces: []
rhel8_interfaces: []

I think that should be done here too.

sanjaychari (changes requested):

sanjaychari (approved):

group_vars/all.yml
rajeshP524: done, added.

PR #541, opened 2025-05-22
Update CentOS repos and modernize shell/Python syntax

imatza-rh:
- Update base container from CentOS 7 to CentOS Stream 8
- Fix CentOS repository mirrors by switching to vault.centos.org
- Replace 'source' with '.' for POSIX shell compatibility
- Update Python venv creation to use 'python3 -m venv' syntax
- Change instackenv.json URLs from /cloud/ to /instack/
- Add certificate validation bypass for scale lab downloads
- Add foreman CLI tools to hammer container


masco:
LGTM!

PR #542, opened 2025-06-18
ocp_on_osp: switch to key-based SSH and set interpreter

imatza-rh:
* Use ansible.builtin.add_host (FQCN).
* Authenticate to undercloud with ansible_ssh_private_key_file instead of ansible_ssh_pass.
* Expose ansible_python_interpreter so RHEL-8 underclouds run modules via /usr/libexec/platform-python.
* Introduce variables undercloud_ssh_key and undercloud_python_interpreter; drop undercloud_password.

This aligns ocp_on_osp.yml with the preceding prepare_* tasks, removes the need for storing clear-text passwords, and prevents "python interpreter not found" errors on first contact.


rajeshP524:
Looks good, but similar changes should go in here: https://github.com/redhat-performance/jetpack/blob/master/ocp_cleanup.yml
could you please make changes in the cleanup file as well?

